PR #5, opened 2021-07-03
Optimised player save data function

steveski:
This addresses the original issue on v1-Final where whilst there was async updates being done, the async list of tasks to run only had a single task anyway so all update statements were still being done one by one. I thought my take on the solution might still be of value on the legacy branch. Rather the run a stack of UPDATE statements which still requires multiple calls across process to MySQL/MariaDB and potentially across the wire if they're not on the same server, then just do a single UPDATE statement for everyone in one step. This is achieved as a set operation across UNIONed SELECT statements representing the value of each column being updated for each player. I've used this approach a couple of times in ESX resources and thread hitches stopped. This level of efficiency takes things to a level where you could save more often without having an adverse effect of the server. You do end up with a larger query to be exexuted but as a "set operation" this is better overall than 64+ individual UPDATE statements being made.

To help with the review, this is a formatted version of the update statement being generated:
```
UPDATE users u JOIN (
  SELECT
    'license:<Some license id>' AS identifier,
    '{"money":94999904,"bank":233091,"black_money":2778349}' AS new_accounts,
    'miner' AS new_job,
    0 AS new_job_grade,
    'superadmin' AS new_group,
    '[]' AS new_loadout,
    '{"x":1779.7,"y":2598.4,"z":45.7,"heading":93.2}' AS new_position,
    '{"repairkit":1,"WEAPON_PISTOL":1,"phone":1,"ammunition_generic":11}' AS new_inventory  

  UNION

  SELECT
      'license:<Some other license id>',
      '{"money":50034,"bank":456542,"black_money":123}',
      'unemployed',
      0,
      'admin',
      '[]',
      '{"x":219.7,"y":598.4,"z":45.7,"heading":45.3}',
      '{"WEAPON_PISTOL":1}'

) vals ON u.identifier = vals.identifier
  SET 
    accounts = new_accounts,
    job = new_job,
    job_grade = new_job_grade,
    `group` = new_group,
    loadout = new_loadout,
    `position` = new_position,
    inventory = new_inventory
```

### DIFF
--- a/[esx]/es_extended/server/functions.lua
+++ b/[esx]/es_extended/server/functions.lua
@@ -196,21 +196,50 @@ ESX.SavePlayer = function(xPlayer, cb)
 end
 
 ESX.SavePlayers = function(cb)
-	local xPlayers, asyncTasks = ESX.GetPlayers(), {}
+	local xPlayers = ESX.GetExtendedPlayers()
+	print("ESX.GetPlayers executed in es_extended saveplayers")
 
-	for i=1, #xPlayers, 1 do
-		table.insert(asyncTasks, function(cb2)
-			local xPlayer = ESX.GetPlayerFromId(xPlayers[i])
-			ESX.SavePlayer(xPlayer, cb2)
-		end)
+	local selectListWithNames = "SELECT '%s' AS identifier, '%s' AS new_accounts, '%s' AS new_job, %s AS new_job_grade, '%s' AS new_group, '%s' AS new_loadout, '%s' AS new_position, '%s' AS new_inventory "
+	local selectListNoNames = "SELECT '%s', '%s', '%s' , %s, '%s', '%s', '%s', '%s' "
+
+	if #xPlayers > 0 then
+		local updateCommand = 'UPDATE users u JOIN ('
+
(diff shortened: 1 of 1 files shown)


ESXCrackhead:
Thank you. I will get this tested.

thelindat:
#### Original method
I found that the ESX.SavePlayers function was triggering 8 saves asynchronously as intended.
Using 200 spoofed xPlayers and an external database yielded the following results:
- ESX.GetPlayers
	- 3 seconds
- ESX.GetExtendedPlayers
	- 3 seconds

Maybe there's some milliseconds of difference occuring in this situation, but as I figured there isn't a need to swap out the function being used.
Repeating the test calling a function from another resource for each player resulted in 12 seconds (22.22ms extra per player)

#### Regarding the new method
After completing the tests of the old save function I went ahead and tested the new save which was instant for the lua side of things, but that doesn't tell us the query time for MySQL. Unfortunately I don't have 200 users in my database, which seems like a hassle to create, so I can't actually test the performance for the query.

How would this bulk update query compare to the one used in esx_status, using case?
https://github.com/esx-framework/esx_status/blob/master/server/main.lua#L86

PR #12, opened 2021-07-28
update de.lua

Konraaaad:
Updated the German Locale of ESX Holdup


nearbyplayer (changes requested):
If a letter is directly behind a quote like with `'Ladenraub:` it should be lowercase, `'ladenraub:`

nearbyplayer:
I was not referring to a singular instance. What I described exists in multiple places, fix it in all those places.

nearbyplayer:
Line 6,9,10 should have remained capital. I said directly after a ‘ only.

nearbyplayer (approved):
Looks good.

PR #15, opened 2021-08-07
Update main.lua

xLamday:
Change at line 424 from 

> Config.EnableDefaultInventory 

to 

> Config.EnableInventory

Create a Config.EnableInventory into config.lua then set it true or false. It depends if the player wants the inventory or maybe not.

If there are errors, tell me pls

### DIFF
--- a/[esx]/es_extended/client/main.lua
+++ b/[esx]/es_extended/client/main.lua
@@ -421,7 +421,7 @@ function StartServerSyncLoops()
 	end)
 end
 
-if Config.EnableDefaultInventory then
+if Config.EnableInventory then
 	RegisterCommand('showinv', function()
 		if not ESX.PlayerData.dead and not ESX.UI.Menu.IsOpen('default', 'es_extended', 'inventory') then
 			ESX.ShowInventory()


[esx]/es_extended/client/main.lua
```diff
@@ -421,7 +421,7 @@ function StartServerSyncLoops()
 	end)
 end
 
-if Config.EnableDefaultInventory then
+if Config.EnableInventory then
```
Comment (FreakOutModz) [inline [esx]/es_extended/client/main.lua:424]: What ? You are chanching a config name for another one ? I don't get it...

FreakOutModz (changes requested):
Why do you have two PR's ? Just make one for both

xLamday:
> Why do you have two PR's ? Just make one for both

I'm new on Github .-.

[esx]/es_extended/client/main.lua
Comment (xLamday) [reply [esx]/es_extended/client/main.lua:424]: Actually yes, because into the config files, there's no Config.Inventory. Instead of going into the code and comment all that  part code, the player can just set it true or false into config.lua , in way to not make the inventory appear

Comment (xLamday) [reply [esx]/es_extended/client/main.lua:424]: > What ? You are chanching a config name for another one ? I don't get it...

I would like to know, if there's some errors now, not if you get it or not. If you don't get it, then just leave other guys see this change, and let em' say whatever they want. And i think you didn't even read the description of this PR, it says the reason of this change.

Comment (thelindat) [reply [esx]/es_extended/client/main.lua:424]: image

You just added an existing config option with a different name.

PR #26, opened 2021-08-20
Created SQL for esx-legacy

DevAnanas0:


Luke20201:
Hi,

We appreciate the Pull Request to help us improve Legacy!
Unfortunately, I will be closing this Pull Request due to the following reasons;

- The SQL is using an outdated version of esx_shops, as this no longer relies on an SQL
- There is not a major use for this for people, not many will want absolutely everything in the repo.
- The SQL file could be formatted better, as of now it is an effective copy-paste of the other SQL files. It would be more fitting to have them do multiple things at once (see how the new ESX_Core.sql add's every column to the users table in one statement, rather than in 10.)

Although this will not be approved, I thank you for your contribution and desire to make ESX a more Developer Friendly and better experience for everyone!

PR #38, opened 2021-10-25
feat: new resource to store health and armour

d3rp-jsx:
Hello, this is a resource I released sometime ago and it's core for RP servers!
Hope this will be useful and feel free to request any edit if needed.

### DIFF
--- a/[esx_addons]/esx_healthnarmour/client/main.lua
+++ b/[esx_addons]/esx_healthnarmour/client/main.lua
@@ -1,0 +1,6 @@
+RegisterNetEvent('esx_healthnarmour:set')
+AddEventHandler('esx_healthnarmour:set', function(health, armour)
+  local playerPed = PlayerPedId()
+  SetEntityHealth(playerPed, health)
+  SetPedArmour(playerPed, armour)
+end)
--- a/[esx_addons]/esx_healthnarmour/esx_healthnarmour.sql
+++ b/[esx_addons]/esx_healthnarmour/esx_healthnarmour.sql
@@ -1,0 +1,3 @@
+ALTER TABLE `users`
+  ADD `health` TINYINT UNSIGNED, 
+  ADD `armour` TINYINT UNSIGNED;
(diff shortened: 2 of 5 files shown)


MrNewb:
Please dont do it this way @d3rp-jsx do it through basic needs so its in the same table. as other information already gets pulled, also eliminates the need for things like ```RegisterNetEvent('esx:onPlayerSpawn')
AddEventHandler('esx:onPlayerSpawn', function()
  local playerId = source
  local xPlayer = ESX.GetPlayerFromId(playerId)

  if xPlayer ~= nil then
    MySQL.Async.fetchAll(LoadHealthNArmour, {xPlayer.identifier}, function(data)
      if data[1].health ~= nil and data[1].armour ~= nil then
        TriggerClientEvent('esx_healthnarmour:set', playerId, data[1].health, data[1].armour)
      end
    end)
  end
end)```

d3rp-jsx:
> Please dont do it this way @d3rp-jsx do it through basic needs so its in the same table. as other information already gets pulled, also eliminates the need for things like ````RegisterNetEvent('esx:onPlayerSpawn') AddEventHandler('esx:onPlayerSpawn', function() local playerId = source local xPlayer = ESX.GetPlayerFromId(playerId)
> 
> if xPlayer ~= nil then MySQL.Async.fetchAll(LoadHealthNArmour, {xPlayer.identifier}, function(data) if data[1].health ~= nil and data[1].armour ~= nil then TriggerClientEvent('esx_healthnarmour:set', playerId, data[1].health, data[1].armour) end end) end end)````

Basic needs is not designed for that, I can't understand what you are saying honestly, two columns won't be any issue.

MrNewb:
by default but its basic design it to save and load a table, why add extra when you already have something doing checks constantly.  my point with the codeblock that doesn't look like a code block is you are doing additional mysql calls when it could be done elsewhere without the need for an additional resource.  Basic needs/status can do this instead.

d3rp-jsx:
You gave yourself the answer, I would never add this to status or basic needs because it doesn't need to be constantly checked and it would be a useless burden to the server.

Luke20201:
Hi,

Thank you for taking the time to create a new resource for ESX! We truly appreciate every commit made to the Repo, in hopes of improving the Framework for everyone!

Unfortunately, however, I believe this will be more fitting as a release on the FiveM Forum's, rather than a part of the ESX repo.

I'm sorry we are unable to approve this, but great work either way!

PR #42, opened 2021-11-01
fix weapon_combatmg_mk2 Locales['en']

etherealth:
remove duplicate Locales['en'] key


Mycroft-Studios (approved):
Seems to fix conflicting keys :)

PR #60, opened 2022-01-16
refactor

thelindat:
Work in progress.

- update to oxmysql
- use prepared statements to improve performance of common queries
- bug fixes
- replace bulk queries from status and esx with prepared statements
- remove async dependency
- remove internal esx variables from the shared table


thelindat:
@Mycroft-Studios @Luke20201 and whoever else, this needs code review and testing.

PR #67, opened 2022-02-05
refactor(taxijob): removing thread's

PsychoShock:
Adding some logic to the code. PLEASE TEST THIS CODE BEFORE MERGE.


PsychoShock:
Logic should be this for all job's. Please use this to make it for every job/script's with thread's.

konijn-x (commented):
Why did you changed pos = {x = .......} to vector3, but size not?

PsychoShock:
Oh lol I forgot this part. Thanks. I will close it and add it in my main PR.

PR #73, opened 2022-02-08
refactor(esx_legacy): Converted all backtick usages to GetHashKey()

GoatG33k:
For FiveM Lua linting compatibility


thelindat:
Fix your Lua linting instead of removing valid CfxLua syntax..

EvanDarwin:
I don't have the time at the moment, feel free to contribute it yourself though!

GoatG33k:
@thelindat Feel free to contribute it yourself, I don't have the time nor willpower to do it. Not to mention, both are valid.

GoatG33k:
> "so it will not break anything"? 

Could you describe what it breaks exactly? To the best of my knowledge there is no functional difference between the two.

Blumlaut:
> To the best of my knowledge there is no functional difference between the two.
GetHashKey is by nature slower. it's also less clean, there is no functional reason to use this compared to just fixing your linter

PR #74, opened 2022-02-08
fix(esx_vehicle): fixed invalid Lua in rental vehicle server-side code

GoatG33k:


DokaDoka:
This was not invalid, esx_vehicleshop has been upgraded to use lua 5.4

DokaDoka:
jesus... 
https://github.com/citizenfx/lua/blob/luaglm-dev/cfx/README.md

GoatG33k:
Yeah, I'm not really a fan of having code that throws errors in the majority of code editors, and isn't supported by the Lua ecosystem... just so you can save like five characters lol

RockySouthpaw:
> Yeah, I'm not really a fan of having code that throws errors in the majority of code editors, and isn't supported by the Lua ecosystem... just so you can save like five characters lol

Welcome to the land of micro optimizations that degrades readability all to reduce some arbitrary **ms** number.

DokaDoka:
whatever, the syntax is valid, if your linter doesn't keep up then that's your problem, and I guess esx's now

PR #75, opened 2022-02-09
refactor(esx_legacy): Converted all GetHashKey to backticks

LukeWasTakenn:
Using backticks is more than 2x faster than using the GetHashKey native (~2.3x to be precise).

https://cookbook.fivem.net/2019/06/23/lua-support-for-compile-time-jenkins-hashes/

This reverts commit 661e21486ea78d639540f23c95b0990d6e994481.


LukeWasTakenn:
Why was this closed?

GoatG33k:
@LukeWasTakenn This is no longer the case, that cookbook entry is from 2019, and there have been performance improvements since that make the performance difference negligible.

LukeWasTakenn:
You do realize backticks runs at compile time and `GetHashKey` runs at runtime, right?

GoatG33k:
Yes. Feel free to actually show a performance impact between the two versions, and I will reconsider my contribution, but you're talking about nanoseconds of overtime.

thelindat:
> nanoseconds of overhead.

Microseconds*
image

But how about, rather than just thinking about performance, we think about actual _good practice_?
If people see ESX swapped from backticks to GetHashKey, well geez it must be _better_ then!

Might as well swap vector distance checks to GetDistanceBetweenCoords.

GoatG33k:
@thelindat Feel free to publish the code you used to generate those results, and I'd be glad to take a look.

>  If people see ESX swapped from backticks to GetHashKey, well geez it must be better then!

Nobody is saying that `GetHashKey()` is superior.

> Might as well swap vector distance checks to GetDistanceBetweenCoords.

[See: slippery slope fallacy](https://en.wikipedia.org/wiki/Slippery_slope)

thelindat:
> cookbook entry is from 2019, and there have been performance improvements since that make the performance difference negligible.

Checked with lmprof, that snippet is twice the speed with backticks.

GoatG33k:
> Checked with lmprof, that snippet is twice as fast with backticks.

If you are using the snippet from the cookbook entry, `GetHashKey()` is invoked 10k times individually in a loop. In the `weapons.config.lua` they are invoked once, ~1,000 times (which using your own numbers is ~4 microseconds). The hash will be calculated at script start, and only calculated once. 

I am still not convinced there is a notable performance impact to the resource(s).

thelindat:
>  The hash will be calculated at script start, and only calculated once.
Thanks for pointing that out, sherlock.

> I am still not convinced there is a notable performance impact to the resource(s).
There isn't a massive performance impact, but it's still there - and for what reason? There's no benefit.

> If they are copying code from esx-legacy then they're not starting in an ideal place anyway...
And now they can copy even worse code.


Weird change for the sake of a "FiveM Lua Linter" that doesn't work with FiveM Lua. You don't have enough time to fix up your linter, but enough to argue that backticks are worthless.

But whatever, you know best and we should ignore the existence of power-patches and extended API that's been added to the language.

GoatG33k:
> There's no benefit.

I think detecting invalid syntax and code quality changes in pull requests is quite useful, as code quality tools have been used in software development workflows for quite a while.

> But whatever, you know best and we should ignore the existence of power-patches and extended API that's been added to the language.

Nope, they exist and are great!

> Thanks for pointing that out, sherlock.

Keep the sarcasm on Discord, you know where to find me 😉

PR #84, opened 2022-02-12
fix identity not showing

sepehr-gr:
add Wait 
good luck :)


PsychoShock:
First, this won't change anything. Also, this was fix in https://github.com/esx-framework/esx-legacy/pull/79

Thanks

PR #86, opened 2022-02-14
fix errors and warnings across esx-legacy

thelindat:


thelindat:
Going to do the inventory stuff separately.

@Mycroft-Studios

Mycroft-Studios:
👍🏼 , will have it tested. looking over the code it seems fine, although, for the Wait Changes, wouldn`t it be better to add sleeps to some of them?

thelindat:
To quote bubble
> Wait will round to the nearest frame (as it’s tick-based) so all you’re doing here is causing problems for anyone with over 200 FPS at no noticeable other gain.
>
> If you want to run something every frame, use 0, don’t try to be clever with values between 1 and 15 or so.

PsychoShock (approved):
Doesn't really change anything so just by looking at it, code will work. It's small improvements. 👍

Mycroft-Studios:
> To quote bubble
> 
> > Wait will round to the nearest frame (as it’s tick-based) so all you’re doing here is causing problems for anyone with over 200 FPS at no noticeable other gain.
> > If you want to run something every frame, use 0, don’t try to be clever with values between 1 and 15 or so.

👍🏼 😄

PR #91, opened 2022-02-16
Fix Bug isDead == 1 it be should isDead == true

NRTnarathip:
The variable type of "isDead" is a boolean, not a number.


Slypher (approved):

Slypher:
I also tested and everything is ok.

Esplike:
Works fine.

PR #110, opened 2022-02-18
fix(sql): added auto-database create

konijn-x:
Creates an database automatic.

### DIFF
--- a/[SQL]/legacy.sql
+++ b/[SQL]/legacy.sql
@@ -1,3 +1,12 @@
+CREATE DATABASE IF NOT EXISTS `esx_legacy`;
+USE `esx-legacy`;
+
+ALTER DATABASE `esx_legacy`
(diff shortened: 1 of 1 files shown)


[SQL]/legacy.sql
```diff
@@ -1,3 +1,12 @@
+CREATE DATABASE IF NOT EXISTS `esx_legacy`;
+USE `esx-legacy`;
```
Comment (PsychoShock) [inline [SQL]/legacy.sql:2]: USE `esx_legacy`;


Fix typo :D

PsychoShock (changes requested):

0xDEMXN (approved):

Mycroft-Studios:
This *may* Break the TxAdmin recipe

Esplike:
I'll have a look into it because I don't want to break the recipe...

Esplike:
This would break the repipe for ESX because TXAdmin automatically creates the database.

PR #112, opened 2022-02-18
fix(locales): update invalid command to fit notifications

0xDEMXN:
Thanks to daddy @thelindat


konijn-x (approved):
Thanks for you contribution! 
❤️The ESX Team

PR #114, opened 2022-02-18
Update functions.lua

sepehr-gr:


Mycroft-Studios:
Will break compatibility with 1. Existing Servers and 2. Multicharacter

PR #149, opened 2022-03-06
Minor Optimization, replaced all table.insert

TheSpaceGamerV2:
Replaced all table.insert by table[#table+1], i didn't do it for the async stuff was too afraid to break something.

### DIFF
--- a/[esx]/cron/server/main.lua
+++ b/[esx]/cron/server/main.lua
@@ -2,11 +2,11 @@ local Jobs = {}
 local LastTime = nil
 
 function RunAt(h, m, cb)
-	table.insert(Jobs, {
-		h  = h,
-		m  = m,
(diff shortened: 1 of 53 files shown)


TheSpaceGamerV2:
For esx_accessories, i initialized vars outside the loop and redefine them inside (macro optimisation) some function was global and this was useless. Server-Side i prevented the error if xPlayer is nil.

TheLazarusRising:
Good work.

But please update this PR with proper commit messages:

TheSpaceGamerV2:
> Good work.
> 
> But please update this PR with proper commit messages:
> 
> https://www.conventionalcommits.org/en/v1.0.0/

Srry, is it good now (i'm new with this) ?

PR #155, opened 2022-03-12
refactor(esx_shops): Add export for OpenShop and make function localized

FjamZoo:
Makes the shops globalized function to a localized function very small change in terms of optimization however its just better code practice.

### DIFF
--- a/[esx_addons]/esx_shops/client/main.lua
+++ b/[esx_addons]/esx_shops/client/main.lua
@@ -1,7 +1,7 @@
 local hasAlreadyEnteredMarker, lastZone
 local currentAction, currentActionMsg, currentActionData = nil, nil, {}
 
-function OpenShopMenu(zone)
+local function OpenShopMenu(zone)
 	local elements = {}
 	for i=1, #Config.Zones[zone].Items, 1 do
 		local item = Config.Zones[zone].Items[i]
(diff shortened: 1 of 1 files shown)


FjamZoo:
Added a export to shop Mycroft told me if you flame me then its his fault not mine

This OpenShop export can be opened by doing this

exports['esx_shop']:OpenShop(Zone) -- The zone is a variable you want to pass to the shop, look inside the shop client file to look more indepth how this function works.

FjamZoo:
for anyone reading this and asking why would we want to make function localised?

localized functions run slightly faster and shows good code practice for new developers looking into how to do stuff.

(Only do this in single files or files where the function is only being accessed in a single file and the local functions have to be above where it gets called like normal Functions)

TheSpaceGamerV2:
> for anyone reading this and asking why would we want to make function localised?
> 
> localized functions run slightly faster and shows good code practice for new developers looking into how to do stuff.
> 
> (Only do this in single files or files where the function is only being accessed in a single file and the local functions have to be above where it gets called like normal Functions)

Sure ur right anyway in this case OpenShopMenu() do not need to be global.

TheLazarusRising:
If you want to finish this PR, then this needs to be done for every resource, along with proper structure for the commit message:

PR #163, opened 2022-03-18
New OnDuty System for jobs.

zf-development:
- Added a setDuty event.
- Added a job.onduty data.
- Default onduty is false at connection and creation of the user, job change.
- Added a esx:toggleDuty event to change the state of the duty.


zf-development:
Cancelled for the moment. PR will be remade when everything will be changed.

PR #164, opened 2022-03-18
feat(es_extended): Added "Duty" system

zf-development:
Thanks Mycroft for the idea of the default state to being true for compatibility... Haven't thought about it.

- setDuty event
- setDuty function
- xPlayer.getJob().job.onduty
- ESX.GetPlayerData().job.onduty

Here's the details.
Translation is needed for all of locales except fr and en.


zf-development:
The spawnpoint config commit was reverted.

[esx]/es_extended/server/main.lua
```diff
@@ -600,3 +601,18 @@ AddEventHandler('txAdmin:events:scheduledRestart', function(eventData)
 		end)
 	end
 end)
+
+RegisterNetEvent('esx:setDuty')
+AddEventHandler('esx:setDuty', function(state)
+    local xPlayer = ESX.GetPlayerFromId(source)
+    if xPlayer.job.onduty == state then return end
+    
+    if state then
```
Comment (PsychoShock) [inline [esx]/es_extended/server/main.lua:610]: I would change that to `bool` to make cleaner. But doesn't really change anything.

[esx]/es_extended/server/common.lua
```diff
@@ -93,6 +93,7 @@ MySQL.ready(function()
 					grade = 0,
 					label = 'Unemployed',
 					salary = 200,
+                    onduty = true,
```
Comment (PsychoShock) [inline [esx]/es_extended/server/common.lua:96]: This need to be `false` no? Since unemployed you are not on dudy.

PsychoShock (changes requested):
Also please use camelcase for a clean code. `onduty` -> `onDuty`

zf-development:
@PsychoShock Code correction have been made.

zf-development:
What do you mean by "true allow by default"? What should be the config?

PsychoShock:
Look good to me, if you wanna be even more helpfull you can add this (for example) in a single job to make people understand how it work.

zf-development:
Yeah sure.

[esx]/es_extended/server/main.lua
```diff
@@ -600,3 +601,18 @@ AddEventHandler('txAdmin:events:scheduledRestart', function(eventData)
...
+    local xPlayer = ESX.GetPlayerFromId(source)
+    if xPlayer.job.onDuty == bool then return end
+    
+    if bool then
+        xPlayer.setDuty(false)
+        xPlayer.triggerEvent('esx:showNotification', _U('stopped_duty'))
+    else
+        xPlayer.setDuty(true)
+        xPlayer.triggerEvent('esx:showNotification', _U('started_duty'))
+    end
```
Comment (PsychoShock) [inline [esx]/es_extended/server/main.lua:616]: This is false since people will use it in this way: `esx:setDuty, true` so `bool` will be true. Code should be replace by:
```lua
if bool then
        xPlayer.setDuty(true)
        xPlayer.triggerEvent('esx:showNotification', _U('started_duty'))
else
        xPlayer.setDuty(false)
        xPlayer.triggerEvent('esx:showNotification', _U('stopped_duty'))
end

PsychoShock (changes requested):
Last change and wil be good.

zf-development:
Done! You'll be able to pull it @PsychoShock.

PsychoShock:
@Mycroft-Studios

TheLazarusRising:
In 'es_extended/server/main.lua' please add the newline at the end of the file.

Esplike:
There is still one more conflict to solve, also if we do add a duty system we better add it into all resources like police and ambulance job etc...

Esplike:
Thanks, i'll add this in as a new thing and we will integrate this into police and ambulance job :)

PR #165, opened 2022-03-19
refactor(esx_legacy): Shorten RegisterNetEvent and AddEventHandler

MonsterTaerAttO:


TheLazarusRising:
Please use proper commit messages:

I should know what you've done without having to look at the code.

[esx_addons]/esx_ambulancejob/fxmanifest.lua
```diff
@@ -4,22 +4,43 @@ game 'gta5'
 
 description 'ESX Ambulance Job'
 
-version '1.6.5'
-
-shared_scripts {
```
Comment (PsychoShock) [inline [esx_addons]/esx_ambulancejob/fxmanifest.lua:9]: huh? Why would you make a change like that...

[esx_addons]/esx_lscustom/client/main.lua
```diff
@@ -380,36 +389,57 @@ CreateThread(function()
...
-		DisableControlAction(2, 170, true)
-		DisableControlAction(2, 167, true)
-		DisableControlAction(2, 168, true)
-		DisableControlAction(2, 23, true)
-		DisableControlAction(0, 75, true)  -- Disable exit vehicle
-		DisableControlAction(27, 75, true) -- Disable exit vehicle
+	while true do
+		Wait(0)
+
+		if lsMenuIsShowed then
```
Comment (PsychoShock) [inline [esx_addons]/esx_lscustom/client/main.lua:432]: also why? Set a function and only call it when needed.

[esx_addons]/esx_mechanicjob/client/main.lua
```diff
@@ -822,28 +818,31 @@ end)
...
 				if v.Type ~= -1 and #(coords - v.Pos) < Config.DrawDistance then
-					Sleep = 0
 					DrawMarker(v.Type, v.Pos.x, v.Pos.y, v.Pos.z, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, v.Size.x, v.Size.y, v.Size.z, v.Color.r, v.Color.g, v.Color.b, 100, false, true, 2, false, nil, nil, false)
 					letSleep = false
 				end
 			end
+
+			if letSleep then
+				Wait(500)
+			end
```
Comment (PsychoShock) [inline [esx_addons]/esx_mechanicjob/client/main.lua:835]: Code like this it's just wrong... please don't use that. Only change the value when needed.

PsychoShock (changes requested):

PR #167, opened 2022-03-20
fix(es_extended/server/main.lua): fix giving weapon components + tints, feat(es_extended/server/main.lua): add distance check

Tinky124:


[esx]/es_extended/server/main.lua
```diff
@@ -360,6 +360,11 @@ if not Config.OxInventory then
 		local playerId = source
 		local sourceXPlayer = ESX.GetPlayerFromId(playerId)
 		local targetXPlayer = ESX.GetPlayerFromId(target)
+        local distance = #(GetEntityCoords(GetPlayerPed(playerId)) - GetEntityCoords(GetPlayerPed(target)))
+        if not sourceXPlayer then return end
+        if not targetXPlayer then print("Cheating: " ..  GetPlayerName(playerId)) return end
+        if distance > 4.0 then print("Cheating: " ..  GetPlayerName(playerId)) return end
```
Comment (PsychoShock) [inline [esx]/es_extended/server/main.lua:366]: Make a config for the distance check since is not everyone that wanna have only `4.0`. Also, since this is a public project, is not everyone that know how to change that :D

PsychoShock (changes requested):

TheLazarusRising:
Please update this PR to use proper commit messages:

TheLazarusRising:
https://github.com/esx-framework/esx-legacy/pull/167#discussion_r832846150

This is accurate, you will need to add config options for some of those variables, as there is no reason why you should force a hardcoded distance.

Tinky124:
@TheLazarusRising here you go man :)

PR #168, opened 2022-03-22
Added support for mf-inventory and mf-multichar.

meta-hub:
Compatibility should not have been broken in any way, in theory...
- Added generic-named events for mundane tasks, allowing multichar and inventory to function without requiring any changes to ESX itself.
- Convert OxInventory compatibility to new "function overwrite" system, allowing multi-inventory support in the future without having to edit `classes/player.lua` code.
- Added MfInventory compatibility via function overwrites.


PsychoShock:
I don't think `ModzFreak` deserve to be in this repo. Most of the people use `ox` now. Will be better to make a fork and add support there. Also, MF doesn't even use Tebex. I recommand ESX to follow FiveM T.O.S. and only support that if they wanna have something related to CFX one day (if happen). My opinion.

Mycroft-Studios:
> I don't think `ModzFreak` deserve to be in this repo. Most of the people use `ox` now. Will be better to make a fork and add support there. Also, MF doesn't even use Tebex. I recommand ESX to follow FiveM T.O.S. and only support that if they wanna have something related to CFX one day (if happen). My opinion.

His Relationship with the ToS is not something that we will be taking into consideration here

TheLazarusRising:
GPL does not allow dynamic linking with incompatible resources/code, so this PR needs some alterations:

1. Any link to mf_inventory must be removed.
2. Any link to any non GPL code must be removed.

TheLazarusRising:
> I don't think `ModzFreak` deserve to be in this repo. Most of the people use `ox` now. Will be better to make a fork and add support there. Also, MF doesn't even use Tebex. I recommand ESX to follow FiveM T.O.S. and only support that if they wanna have something related to CFX one day (if happen). My opinion.

Thank you for your opinion, ESX as an organization is not involved with selling in any form, so what you're saying does not make sense.

"If they wanna have something related to CFX one day"? What does this even mean? We don't want to be Cfx, we have been fine since 2017, and we will continue to be fine following our own ruleset.

PR #175, opened 2022-03-27
refactor(esx_lscustom/client/main.lua): fixed loop

iSentrie:
Loop now in affect and doesn't crash, because of missing `Wait(0)`


TheLazarusRising:
Please update this PR title using:

iSentrie:
> 

the hell is that, i don't know what is that or how to use that.

TheLazarusRising:
That would be proper commit messages on a repo.

In this PR, you simply moved code around, without changing functionality, so the proper title for this PR would be:

`refactor(ls_custom/client/main.lua): Fixed loop`

TheLazarusRising:
Thank you.

PR #189, opened 2022-04-02
add bt-target support for esx_sit

someone0002:
If those who prefer to use bt-target instead of using a command to sit-down in a chair, this will offer that support.

### DIFF
--- a/[esx_addons]/esx_sit/client.lua
+++ b/[esx_addons]/esx_sit/client.lua
@@ -145,3 +145,24 @@ function sit(object, modelName, data)
 		end
 	end, objectCoords)
 end
+
+Citizen.CreateThread(function()
+	local Sitables = {}
+
+	for k,v in pairs(Config.Interactables) do
+		local model = GetHashKey(v)
+		table.insert(Sitables, model)
+	end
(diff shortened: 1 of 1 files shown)


someone0002:
sit:sit might be the wrong client trigger to sit down in the chair.

TheLazarusRising:
Please update this PR to use proper commit messages:

You can look in #news in ESX Discord to get more information if you need to.
Once the title is corrected, then it can be reviewed.

[esx_addons]/esx_sit/client.lua
```diff
@@ -144,4 +144,25 @@ function sit(object, modelName, data)
...
+
+Citizen.CreateThread(function()
+	local Sitables = {}
+
+	for k,v in pairs(Config.Interactables) do
+		local model = GetHashKey(v)
+		table.insert(Sitables, model)
+	end
+	Wait(100)
+	exports['bt-target']:AddTargetModel(Sitables, {
```
Comment (Mycroft-Studios) [inline [esx_addons]/esx_sit/client.lua:157]: this will error if you dont use bt-target, please make it a config option :)

someone0002:
Will update then create another PR

PR #190, opened 2022-04-03
Tweak: Jenkins Hashes

FantomasIsBad:
Hello, I turned almost all GetHashKeys into Jenkins Hashes...


iSentrie:
That was done before, but for some reason was reverted.

TheLazarusRising:
The backticks/jenkins were removed to enable a linter on the Github. While I may not necessarily agree with backticks over GetHashKey or joaat, we aren't going to continue going back and forth on this.

PR #191, opened 2022-04-03
refactor(es_extended/server/commands.lua): Alter players command to use `ESX.GetExtendedPlayers()` rather than `ESX.GetPlayers()`

FantomasIsBad:


TheLazarusRising:
In the future, you need to title your PR properly, otherwise, it will sit here without being reviewed until you make the changes.

FantomasIsBad:
Sorry :D

PR #206, opened 2022-04-11
Add esx_shops.sql. This table is being used in esx_optionalneeds but …

manutrix:
…never created


Mycroft-Studios:
Shouldnt have been merged, useless SQL since nothing uses that table.

manutrix:
If you look at esx_optionalneeds, it uses that table. The sql script of esx_optionalneeds fails because that table doesn't exists

PR #207, opened 2022-04-12
Add shops table

manutrix:
The table `shops` doesn't exist in any other module. I've created [a pull request to add it in esx_shops](https://github.com/esx-framework/esx-legacy/pull/206) but it has been rejected. So, even if it doesn't make sense for me, let add it in the module that uses it.

### DIFF
--- a/[esx_addons]/esx_optionalneeds/esx_optionalneeds.sql
+++ b/[esx_addons]/esx_optionalneeds/esx_optionalneeds.sql
@@ -1,5 +1,14 @@
 USE `es_extended`;
 
+CREATE TABLE `shops` (
+	`id` int(11) NOT NULL AUTO_INCREMENT,
+	`store` varchar(100) NOT NULL,
+	`item` varchar(100) NOT NULL,
+	`price` int(11) NOT NULL,
+
+	PRIMARY KEY (`id`)
+);
+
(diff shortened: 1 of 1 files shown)


manutrix:
Il look at the details and don't see the error. Something I can do?

Mycroft-Studios:
ESX shops does not use a SQL table so why would u add one? makes 0 sense.

manutrix:
Okay. But why do you have:

```
INSERT INTO `shops` (store, item, price) VALUES
	('TwentyFourSeven', 'beer', 45),
	('RobsLiquor', 'beer', 45),
	('LTDgasoline', 'beer', 45)
;
```

Into that sql file then?!

Mycroft-Studios:
> @Mycroft-Studios,
> 
> Okay. But why do you have:
> 
> ```
> INSERT INTO `shops` (store, item, price) VALUES
> 	('TwentyFourSeven', 'beer', 45),
> 	('RobsLiquor', 'beer', 45),
> 	('LTDgasoline', 'beer', 45)
> ;
> ```
> 
> Into that sql file then?!
> 
> I will then create another PR to remove that insert!

feel free to pr the removal of that, that was probably overlooked when making the legacy.sql 🙃

PR #214, opened 2022-04-19
feat(es_extended/client/functions.lua): Added drift tires natives

Deliux:
Introduced in v2372 tuner dlc build. Possible use with ls lustoms script to expand tuning possibilities.


iSentrie (approved):

Benzo00:
Dont see the point in this.

iSentrie:
It's worth having, thanks Deliux. I don't see point for not merging it, non sense.

PR #216, opened 2022-04-20
Feat(es_extended/client/functions.lua): Add RGB saving to vehicles

tr1ckz:
Add RGB support and fixed liveries


iSentrie:
Tested RGB ✅

Slypher (approved):

PR #220, opened 2022-04-25
refractor(loadingscreen): updated client files for effect

douth-c:


douth-c:
Locally tested across two clients.

Bugs: single tick of white space between client and client fade out. Nothing appears to fix it.

PR #247, opened 2022-05-07
Fix - LS Custom wheels changement on bikes

Slewog:
Change the position of the condition GetVehicleClass (not working before and working now)


Slewog:
Hi Mycroft, I saw that the ls Custom bug was not yet fixed so I created commits so that you only have to validate it

PR #249, opened 2022-05-11
fix: Fix everytime player join server found dead if he was alive at time of leaving server

mr-capy:
before somehow dev remove the check so i just put back the check to get status if player is dead according to saved values in database then on player join it will not kill player but if player is dead according to database it will kill player on server join 
its a bug in this release so i fix it back and copy this code from old files of legacy not my own written.


mr-capy:
if we remove this line which i add from old files then on player join ambulance script always kill player either in database isDead is 0 or 1

thelindat:
This is a breaking change, fix your database instead; the datatype for `is_dead` should be TINYINT(1), which is typecast as a boolean value (true or false) by both oxmysql and mysql-async. The issue with the datatype is a result of issues with zap's ESX pack.

For an actual solution that works for everybody, check if the value is 0.
```lua
cb(isDead == true or isDead == 1)
```

mr-capy:
Ok sir i am closing this request

Issue #166, opened 2022-03-20
Issue with esx_vehicleshop


mr-capy:
Config.EnablePlayerManagement     = true
then when pop vehicle or put out vehicle for sale then this vehicle name always nill i tested it on ingame vehicles like blista and etc
i am also providing screenshot
![Screenshot (103)](https://user-images.githubusercontent.com/78591839/159157431-f664cbb4-dcb0-4cff-a611-339cb9635b61.png)

LEGITIRONY:
Have you modified this esx_vehicleshop and have you imported the sql file correctly, also are you getting any errors in your F8 or console

mr-capy:
Didn't modify and also correctly imported sql everything else is fine when purchasing vehicle but in this menu it shows nill

Bluxrs:
Add the name and model into the database, see what it returns then. If it still does not show then probably a bug but I doubt it as it looks in order in the client and server.lua files for getting vehicle label

mr-capy:
> Add the name and model into the database, see what it returns then. If it still does not show then probably a bug but I doubt it as it looks in order in the client and server.lua files for getting vehicle label

I was using the ingame vehicles which are in vehicles table also with model price etc but still same

CsokiHUN:
I fixed this bug, check it out. [https://github.com/esx-framework/esx-legacy/pull/250](url)

PR #250, opened 2022-05-11
Fixes issue #166 + small optimalize

CsokiHUN:
I fixed bug #166, check it out, it's been successful.

### DIFF
--- a/[esx_addons]/esx_vehicleshop/server/main.lua
+++ b/[esx_addons]/esx_vehicleshop/server/main.lua
@@ -25,7 +25,7 @@ end)
 
 function SQLVehiclesAndCategories()
 	categories = MySQL.query.await('SELECT * FROM vehicle_categories')
-	vehicles = MySQL.query.await('SELECT vehicles.*, vehicle_categories.label AS categoryLabel FROM vehicles, vehicle_categories WHERE vehicles.category = vehicle_categories.name')
+	vehicles = MySQL.query.await('SELECT vehicles.*, vehicle_categories.label AS categoryLabel FROM vehicles JOIN vehicle_categories ON vehicles.category = vehicle_categories.name')
 end
 
 function getVehicleFromModel(model)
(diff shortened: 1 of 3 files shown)


CsokiHUN:
If the script is set to be handled by the faction, no vehicles were available in the menu. that's why i wrote the `getVehicles()` function
I put a while instead of an event because any script will restart. But I didn't know how to break the multicharacter

CsokiHUN:
I added a new thread that only runs when the player has logged in. `getVehicles()` is needed because if the client starts earlier than the server, it will not send out the vehicles, categories tables. I hope this is okay.

CsokiHUN:
i fixed, check it please

[esx_addons]/esx_vehicleshop/server/main.lua
```diff
@@ -25,26 +25,7 @@ end)
...
-				vehicle.categoryLabel = category.label
-				break
-			end
-		end
-	end
-
-	-- send information after db has loaded, making sure everyone gets vehicle information
-	TriggerClientEvent('esx_vehicleshop:sendCategories', -1, categories)
-	TriggerClientEvent('esx_vehicleshop:sendVehicles', -1, vehicles)
+	vehicles = MySQL.query.await('SELECT vehicles.*, vehicle_categories.label AS categoryLabel FROM vehicles, vehicle_categories WHERE vehicles.category = vehicle_categories.name')
```
Comment (boermansjo) [inline [esx_addons]/esx_vehicleshop/server/main.lua:28]: Why not start by using actual joins in queries ?
`SELECT vehicles.*, vehicle_categories.label AS categoryLabel FROM vehicles JOIN vehicle_categories ON vehicles.category = vehicle_categories.name`

Here are some guides:
JOIN data in ON
Filter data in WHERE
Write explicit JOINs to make your Query more readable
Filter data in the WHERE clause instead of the JOIN to ensure it is correct and readable
Different SQL languages may have different query plans based on filtering in the ON clause vs the WHERE clause, so test the performance on your database

Comment (CsokiHUN) [reply [esx_addons]/esx_vehicleshop/server/main.lua]: You're right. I replaced.
I didn’t notice much difference in performance.

Comment (boermansjo) [reply [esx_addons]/esx_vehicleshop/server/main.lua]: There is no performance issues. It is just that a WHERE is for filtering data.
In SQL it is always best practice to use explicit joins and it is more readable.

Comment (CsokiHUN) [reply [esx_addons]/esx_vehicleshop/server/main.lua]: I understand. I’m not the best at more complex SQL queries yet but I’m trying. Thanks!

PR #257, opened 2022-05-14
feat(locales): Added slovenian locales

PandaSLO:
Added slovenian locales to all ESX Addon scripts.


Benzo00:
you are removing "en" translation, go over your files and fix it before we can accept

PR #261, opened 2022-05-15
feat(locales) Slovenian translation

PandaSLO:


Benzo00:
For those that are trying to contribute to ESX.

Please follow this, as your PRs will not be accepted nor reviewed until your PR titles are following this properly.

1. refactor()
Refactoring is when you change the code, without changing functionality. So minor fixes where the code does the same thing in the end.

2. feat()
feat is for all new code.

3. fix()
fix is just for fixing mistype, or broken code

4. feat!()
all new code, that introduces breaking changes

5. fix!()
fix broken code, but introduce breaking changes

PR #267, opened 2022-05-21
fixed the positioning for top-right

xorzo1:
fixed the positioning for top-right because it wasen't the same as top-left xD

Preview:
https://imgur.com/a/mGJdrLB


PandaSLO:
30 secs ago

PR #291, opened 2022-06-04
feat(server.cfg): Added 2545 gamebuild.

PandaSLO:


iSentrie:
This one is totally broken version, use at least 2612

PR #302, opened 2022-06-07
refactor(esx_menu_default/dialog): Optimalization

itIsMaku:
Optimized checking keys and actions.
Removed while true do loops.

### DIFF
--- a/[esx]/esx_menu_default/client/main.lua
+++ b/[esx]/esx_menu_default/client/main.lua
@@ -1,8 +1,13 @@
 local GUI, MenuType = {}, 'default'
 GUI.Time = 0
+local OpenedMenus = 0
 
 local function openMenu(namespace, name, data)
-	SendNUIMessage({
+	OpenedMenus = OpenedMenus + 1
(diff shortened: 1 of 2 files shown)


[esx]/esx_menu_default/client/main.lua
```diff
@@ -1,93 +1,116 @@
-local GUI, MenuType = {}, 'default'
+local GUI, MenuType = {}, "default"
```
Comment (mcNuggets1) [inline [esx]/esx_menu_default/client/main.lua:1]: very optimised

Comment (itIsMaku) [reply [esx]/esx_menu_default/client/main.lua:1]: You should take a look on commits. As you can see, there is a commit with reformatting.

Comment (mcNuggets1) [reply [esx]/esx_menu_default/client/main.lua:1]: But why in a different stlye than the rest of ESX Legacy, doesnt make sense

GRZ-D3V (approved):

Benzo00:
Proper title is needed, check issues here on legacy GitHub to see pr guidelines

PR #304, opened 2022-06-08
Optimizing menu_default

LABNetwork1:
Removed all loops


itIsMaku:
I’ve tried this but it didn’t work for me. Keys wasn’t detected. 
I did it already there: https://github.com/esx-framework/esx-legacy/pull/302

PR #321, opened 2022-06-26
Fix esx_notify not supporting notification color code

xrb936:


iSentrie:
⁉️   It should not work like that anyway.

PR #329, opened 2022-07-04
refactor(Changed fx_version from adamant to cerulean + added lua 54)

ItsManueh:
Change fx_version adamant to use cerulean (fxv2)
and added lua54 in a couple of scripts

Replace NUI http for https


Benzo00:
You can see it here https://github.com/esx-framework/esx-legacy/issues/343

ItsManueh:
> You can see it here #343

Okay thanks :D

PR #345, opened 2022-07-17
added support for without options

AamiRobin:
I just added support for without options so you can just use the Progressbar.


Benzo00:
https://github.com/esx-framework/esx-legacy/issues/343
check this.

AamiRobin:
is it going to break the system if you add this or do you not want to add it? and I know we can use notify.

PR #346, opened 2022-07-17
better import

BaziForYou:
- auto job cache if use import

### DIFF
--- a/[esx]/es_extended/imports.lua
+++ b/[esx]/es_extended/imports.lua
@@ -19,4 +19,9 @@ if not IsDuplicityVersion() then -- Only register this event for the client
         ESX.PlayerLoaded = false
         ESX.PlayerData = {}
     end)
+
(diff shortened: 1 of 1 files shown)


mahanmoulaei (commented):
based on what I have tested so far if you load imports.lua in fxmanifest, you don't need an event handler for setJob in client-side anymore since the whole job table values get updated at line [#L6](https://github.com/esx-framework/esx-legacy/blob/7bdefc289480b9f67e8feb30bc20f8146a98a417/%5Besx%5D/es_extended/imports.lua#L6)
```lua
ESX.PlayerData[key] = val
```
You can also listen to the PlayerData changes by its key through OnPlayerData at line [#8](https://github.com/esx-framework/esx-legacy/blob/7bdefc289480b9f67e8feb30bc20f8146a98a417/%5Besx%5D/es_extended/imports.lua#L8)
```lua
OnPlayerData(key, val, last)
```

PR #348, opened 2022-07-18
added self.setDuty to xPlayer class

ITKewai:
https://github.com/esx-framework/esx-legacy/blob/f2671b0ab6574467012572e814d944e3b76088bb/%5Besx%5D/es_extended/server/main.lua#L689


badfantomasss:
You can see it here https://github.com/esx-framework/esx-legacy/issues/343

Benzo00:
Duty system is getting removed/redone so no point in this pr

PR #360, opened 2022-07-26
Update miner.lua

progamernoob:
fix elevation of marker.


Benzo00:
Esx jobs is getting a full remake so will not add an thing to the old one

PR #390, opened 2022-08-11
improve(client/main.lua): make sure this event dont call in client side

NRTnarathip:
When you want register net event trigger by server side only
you can use
RegisterNetEvent(eventName, function()
    print("hello")
end)

**should not be used**
RegisterNetEvent(eventName)
AddEventHander(eventName, function()
   print("hello world")
end)

image
other event. i recommanded refector, improve code if you have free time
Thanks a lot.😁

### DIFF
--- a/[esx]/es_extended/client/main.lua
+++ b/[esx]/es_extended/client/main.lua
@@ -13,8 +13,7 @@ CreateThread(function()
 	end
 end)
 
-RegisterNetEvent('esx:playerLoaded')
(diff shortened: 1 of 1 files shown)


Mycroft-Studios:
other way around, 
its RegisterNetEvent if u want it to be on both

NRTnarathip:
Ohhhhh my godd
I'm never know oh shit 
Thanks bro for infomation 😅
last one question
why not use 
RegisterNetEvent 
it write to easy and fast

mcNuggets1:
Would that change so much?

Issue #370, opened 2022-07-29
es_extended - throwables, parachutes and gas canisters are not removed by the server


mcNuggets1:
**Describe the bug**
Parachutes, Gasoline canisters, Throwables etc. aren't removed by the server from the player, when using them all out.
You could just reconnect or relog and have atleast 1 of them again.
Works for parachutes in a way that grants infinite parachutes.

**To Reproduce**
1. Have a parachute
2. Use it
3. See it's gone
4. Rejoin
5. It's back

**Expected behavior**
Misc items getting removed by the server after reaching the end of them.

**Screenshots**
None needed

**Debug Info (please complete the following information):**
Irrelevant

**Additional context**
Not needed

michaeld256:
Fixed in #394

iSentrie:
that fix was rejected?

mcNuggets1:
It wasn't finished.
But no comment from anyone tho.

michaeld256:
Yea I was waiting for maintainers feedback on which path to go before commiting the followup solution regarding the possible cheater abuse discussed in the PR.
Got closed for whatever reason though.

Gellipapa:
Hi! @mcNuggets1 Thank you for create issue.  I honestly don't know why this issue was closed, but I'm here now and we can talk about it. 

If I understand correctly, the problem is that certain items are not deleted by the server when they are used up?

What is now the status of this fix? 

Thank you for answers.

mcNuggets1:
Items are not removed and the fix was closed

Gellipapa:
@mcNuggets1  Why must it always run? Can't we check if someone has thrown something away/used it up? What do you think about this?

mcNuggets1:
I honestly don't know any game events that do alert us of that event.

Gellipapa:
I see. 
```
			CreateThread(function()
				while ESX.PlayerLoaded do
					for i, weapon in ipairs(ESX.GetPlayerData().loadout) do
						local playerHasWeapon = HasPedGotWeapon(ESX.GetPlayerData().ped, GetHashKey(weapon.name), false)

						if not playerHasWeapon then
							TriggerServerEvent('esx:removeWeapon', weapon.name)
						end
					end 


					Wait(1000)
				end
			end)
```

That's what I meant. That's always the run for cycle but really you should only be holding a gun shouldn't you?

mcNuggets1:
I didn't create the code.
Optimising the code would look the following:
[...]

PR #394, opened 2022-08-12
Fix(esx/main): throwables, parachutes and gas canisters are not removed by the server

michaeld256:
(Issue #370)

The server persists incorrect client loadouts whenever a weapon was removed by native GTA features, such as using your parachute. GTA removes certain weapons whenever they're used. The server is not getting informed about that removal.

To fix this issue, the client needs to be synchronized with the servers state about it's loadout, so it is able to periodically check if the items that are supposed to be there are actually there. If not, due to e.g. GTA removed a weapon, the server will be informed and the players server-loadout will be adjusted accordingly

### DIFF
--- a/[esx]/es_extended/client/main.lua
+++ b/[esx]/es_extended/client/main.lua
@@ -421,8 +421,23 @@ end
 
 function StartServerSyncLoops()
 	if not Config.OxInventory then
-			-- keep track of ammo
+			-- keep track of weapons in loadout
+			CreateThread(function()
+				while ESX.PlayerLoaded do
+					for i, weapon in ipairs(ESX.GetPlayerData().loadout) do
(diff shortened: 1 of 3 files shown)


mcNuggets1:
What if a cheater removed all of your weapons, which is possible?

michaeld256:
> What if a cheater removed all of your weapons, which is possible?

Being honest I haven't really been considering cheaters to remove items which would lead to persisting cheaters actions.
Though I feel like having the server persist the actual correct state of the player is the better solution here, let there be a cheater or not. 
The consequence of not fixing this issue is an infinite stock of said items, though reconnecting after each use is required. 


Maybe a good trade off to this problem would be something like a whitelist for only certain weapons (such as parachutes, throwables, gas canisters etc.) to be affected by this fix/PR.

mcNuggets1:
Is there a reason this was closed?

iSentrie:
this removes all weapons you have when you join the server

PR #422, opened 2022-08-27
refactor: esx_vehicleshop/client/main.lua esx:setJob fixed

mr-capy:
some how the old code does not detecting the job changings of cardealer job when Config.EnablePlayerManagement  = true


mr-capy:
still buging in some situations

PR #424, opened 2022-08-28
Update functions.lua

Gasparpereira2004:
ESX.Game.GetVehicleProperties is not getting livery correctly, now fixed with GetVehicleLivery(vehicle)

### DIFF
--- a/[esx]/es_extended/client/functions.lua
+++ b/[esx]/es_extended/client/functions.lua
@@ -823,7 +823,7 @@ function ESX.Game.GetVehicleProperties(vehicle)
             modTrimB = GetVehicleMod(vehicle, 44),
             modTank = GetVehicleMod(vehicle, 45),
             modDoorR = GetVehicleMod(vehicle, 47),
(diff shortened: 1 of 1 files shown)


thelindat:
@Mycroft-Studios 

Just so you know, vanilla vehicles (maybe some DLC and custom) use GetVehicleLivery; otherwise newer vehicles use GetVehicleMod. https://github.com/esx-framework/esx-legacy/commit/7dbbe28a5340e508b6d22bd96d73c2d2a305105c was a breaking change.

```lua
local modLivery = GetVehicleMod(vehicle, 48)

if modLivery == -1 then
    local modLivery2 = GetVehicleLivery(vehicle)

    if modLivery2 ~= 0 then
        modLivery = modLivery2
    end
end
```
```lua
if props.modLivery then
    SetVehicleMod(vehicle, 48, props.modLivery, false)
    SetVehicleLivery(vehicle, props.modLivery)
end
```

PR #428, opened 2022-09-01
Fix Error Create Missing Pickup

Rungrot444:


iSentrie:
with your fix, old issue comes up again

https://github.com/esx-framework/esx-legacy/issues/204

which was never fixed, at least from my tests.

Benzo00:
https://github.com/esx-framework/esx-legacy/pulls take a look in this

PR #432, opened 2022-09-02
tweak(es_extended/client/main.lua): more optimization

mahanmoulaei:
- Changed the result/behavior of Config.DisableNPCDrops which disables all pickups now with no loop needed.

- Re-assigned a longer wait time for the pause menu thread that modifies the HUD display.


mcNuggets1:
Does look like a mess

Benzo00:
closing this now due to conflicts.

PR #433, opened 2022-09-02
esx_clotheshop Replaced Helptext with TextUI

ofcshiro:
Replaced the Helptext with TextUI 

Clotheshop runs with that 0.03 - 0.04ms if you see / be inside of a Marker


Benzo00:
will redo this!

PR #436, opened 2022-09-03
fix major issue in esx_jobs

mindkind:


Benzo00:
esx jobs is being fully redone and replaced very soon so we will not bother accepting anything todo with the old one!

PR #442, opened 2022-09-05
feat: (created Italian language for esx_banking)

bitpredator:


Benzo00:
Will be a new esx_banking so this will be closed

PR #450, opened 2022-09-06
feat(state bag/handcuff): add player state

GRZ-D3V:


GRZ-D3V:
Also note that server will not replicate on client since there's no need to.

GRZ-D3V:
Thanks @xrtw

PR #458, opened 2022-09-10
Create nl.lua

HariboInc:


Benzo00:
image
your overlaping the en translation, also use a proper title

PR #468, opened 2022-09-15
fix(esx_property): duplicated furniture  

ItzTrivial:


Gellipapa:
Hi! @ItzTrivial Please use default PR format. Change PR title to fix(duplicated props in config.lua in esx_property). 

Thank you for help remove these props 👍

PR #471, opened 2022-09-16
fix(es_extended/server/onesync.lua): fixed ESX.OneSync.SpawnVehicle

vrnagy:
State set failed because the spawned car's variable name overwritten the Entity function


Gellipapa:
Hi! Thank you for fixed. 
Soon @Benzo00  will accept.

Benzo00:
should already be fixed

PR #475, opened 2022-09-17
refactor(esx_status)

CsokiHUN:


CsokiHUN:
Hello @Mycroft-Studios! Any reason for close pr?

Mycroft-Studios:
would break too many resources

CsokiHUN:
> would break too many resources

what?? The events same as in older solutions..
It will not load the status from the old data for the first time, but I think that is negligible.

PR #491, opened 2022-09-21
refactor(esx_weaponshop): Dutch Translations

HariboInc:


HariboInc:
I actually see people doing feat but a translation is just a refactor right, because we ain't changing a thing except a few words.

PR #519, opened 2022-10-01
refactor(esx_taxijob): Dutch translations

HariboInc:


Benzo00:
esx_taxijob is getting a rewrite so will not accept any prs to it, the new one will be in very soon

PR #521, opened 2022-10-01
refactor(esx_phone): Dutch translations

HariboInc:


Benzo00:
will accept this, but this phone will soon get removed.

PR #524, opened 2022-10-01
refactor(esx_jobs): dutch sql

HariboInc:


Benzo00:
esx_jobs is getting a full rewrite so i will not accept any prs for it rn

PR #529, opened 2022-10-02
fix(esx_banking): set ox_target correctly

FAr18:
1. Set ox_target with correct parameter format.
2. ox_target uses name as identifier instead of label when removing option.


Benzo00:
Closing this due to ox target will be removed cuz we have our own plans ;)

PR #531, opened 2022-10-02
refactor(esx_addons): Dutch Translations

HariboInc:
Srry for all the loose pull requests, i'm a github noob :P I'll try to do it like this whenever.


HariboInc:
Yo, don't want to bug but what's up?

tomiichx:
I wouldn't call it a 'refactor' if you're adding new translations..

HariboInc:
1. refactor()
Refactoring is when you change the code, without changing functionality. So minor fixes where the code does the same thing in the end.

Translations doesnt change the code in any way. Its just simple text that is different. Script functionality doesnt change.

tomiichx:
Sure, also makes sense.. But adding something new is a **feature**.
Please take a look [here](https://www.conventionalcommits.org/en/v1.0.0-beta.2/)

HariboInc:
Appreciated but my other refactor translations have been merged.

I'm not feating anything new except a Hello for Hallo. That is refactoring in my eyes. Because I am not editing nor adding code lines.

PR #532, opened 2022-10-02
Danish ES_extented translation

BreexDK:


Benzo00:
first of all, 
image
your overwriting the en translation and second this isnt the newest translations so please update your es_extended and fix the overwriting en issue

PR #534, opened 2022-10-02
fix(esx_policejob) - adding missing title translation

Deliux:


Benzo00:
if u want to add this u need to add it to all translations, otherwise it will say "translation doesnt exists" for everyone that has anything else than en

Deliux:
Don't really know all these languages, is just using translator acceptable?:D

Benzo00:
aka add this ['police_actions'] = 'police actions', in all translations

PR #539, opened 2022-10-03
feat:(Italian translation)

bitpredator:


Benzo00:
taxijob is getting a full rewrite so wont accept any prs due to changes it will have

PR #543, opened 2022-10-05
esx society > esx context

Taavi-AU:
-- Added esx context support
-- Added check balance option to see how much money inside society account
-- Added grade label change menu


Taavi-AU:
tested and works perfectly

still uses esx dialog menu, since couldnt figure out the inputs in context menu 🤣

Taavi-AU:
latest commits is esx context support for the following resources;
-- esx_bankerjob
-- esx_barbershop
-- esx_clotheshop
-- esx_dmvschool

PR #546, opened 2022-10-06
feat(es_extended/classes/player.lua): add esx:setGroup event

CsokiHUN:
Title has everything
On the server side, xPlayer can be used instead of serverId, I am waiting for suggestions.

### Server
```lua
AddEventHandler("esx:setGroup", function(serverId, group, oldGroup)
print(serverId, group, oldGroup)
end) 
```
### Client
```lua
RegisterNetEvent("esx:setGroup", function(group, oldGroup)
print(group, oldGroup)
end)
```

### DIFF
--- a/[esx]/es_extended/server/classes/player.lua
+++ b/[esx]/es_extended/server/classes/player.lua
@@ -75,9 +75,14 @@ function CreateExtendedPlayer(playerId, identifier, group, accounts, inventory, 
 	end
 
 	function self.setGroup(newGroup)
+		local old = self.group
+
 		ExecuteCommand(('remove_principal identifier.%s group.%s'):format(self.license, self.group))
 		self.group = newGroup
 		ExecuteCommand(('add_principal identifier.%s group.%s'):format(self.license, self.group))
+
+		self.triggerEvent('esx:setGroup', newGroup, old)
+		TriggerEvent('esx:setGroup', self.source, newGroup, old)
 	end
 
 	function self.getGroup()


Mycroft-Studios:
whatttt? why is this even a thing?

CsokiHUN:
> whatttt? why is this even a thing?

We can use it in external scripts to use the change of the group.

Mycroft-Studios:
> > whatttt? why is this even a thing?
> 
> We can use it in external scripts to use the change of the group.

that is literally useless.

Mycroft-Studios:
all this adds is shit loads of security issues :)

CsokiHUN:
> all this adds is shit loads of security issues :)

this does not change the group, it only runs when it changes.

> that is literally useless.

You can log in to discord if your admin's group changes, for example

CsokiHUN:
If ox_core includes it, why can't esx do this?

Mycroft-Studios:
> > all this adds is shit loads of security issues :)
> 
> this does not change the group, it only runs when it changes.
> 
> > that is literally useless.
> 
> You can log in to discord if your admin's group changes, for example
Mod menus will eat this event like its breakfast, there is literally 0 validation here and the client shouldn't ever be able to have full access over something as important as your group since the clients can manipulate that all they want, if u want group, u should always use a server callback or just use xPlayer.group

CsokiHUN:
This event still does not change any data!
But if the client event is such a big problem, it should only be on the server side.
I hope you at least looked at the code because it looks like you didn't..

CsokiHUN:
Then please tell us how to check that the player's group changes. I don't want to run a thread unnecessarily.

Mycroft-Studios:
> Then please tell us how to check that the player's group changes. I don't want to run a thread unnecessarily.

give me 1 valid reason why the client needs to know that

CsokiHUN:
client is not important.
I can't check on the server if the player's group changes.
I would like to log on to discord to see if the group changes. to solve it so that the server machine does not burn because I am running an unnecessary thread.

Mycroft-Studios:
or just ya know, add a log to the command, adding a event that will allow modders potential access to admin only shit, **VERY** easily , is not worth it just to have something log to discord, especially since ESX literally now has a discord log feature built into it

CsokiHUN:
## This event does not change the data! 
### If the client-side event is deleted, modders have no chance either.
It only returns with the data so that we know what has changed.

Mycroft-Studios:
there is still 0 point, this is just havent an event there for the sake of having an event there

CsokiHUN:
Why not add new features?
Maybe someone needs it like I do. I don't think that one event causes so much performance degradation

Mycroft-Studios:
because nobody needs this. your 1 situation for it, is completely invalid since ESX has functions u can use to do that, u dont need an event for this

CsokiHUN:
What function are you talking about?
You still haven't described how I could solve my problem by wasting unnecessary performance.

Mycroft-Studios:
> What function are you talking about? You still haven't described how I could solve my problem by wasting unnecessary performance.

[ESX.DiscordLog](https://github.com/esx-framework/esx-legacy/blob/main/%5Besx%5D/es_extended/server/functions.lua#L268-L314)

sadam376:
Bohóc microsoft

PR #547, opened 2022-10-06
Updater

zRxnx:
[+] Updater

(Maybe its needed)

### DIFF
--- a/[esx]/es_extended/server/updater.lua
+++ b/[esx]/es_extended/server/updater.lua
@@ -1,0 +1,57 @@
+local curVersion = GetResourceMetadata(GetCurrentResourceName(), "version")
+local resourceName = "es_extended"
+
+if Config.CheckForUpdates then
+    CreateThread(function()
+        if GetCurrentResourceName() ~= "es_extended" then
+            resourceName = "es_extended (" .. GetCurrentResourceName() .. ")"
+        end
+    end)
+
+    CreateThread(function()
+        while true do
+            PerformHttpRequest("https://api.github.com/repos/esx-framework/esx-legacy/releases/latest", checkVersion, "GET")
+            Wait(3600000)
+        end
+    end)
+
+    function checkVersion(err, responseText, headers)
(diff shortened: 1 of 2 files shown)


[esx]/es_extended/server/updater.lua
```diff
@@ -0,0 +1,57 @@
...
+                repoURL = "https://github.com/esx-framework/esx-legacy"
+            end
+        end, "GET")
+
+        repeat
+            Wait(50)
+        until (repoVersion and repoURL and repoBody)
+
+        return repoVersion, repoURL, repoBody
+    end
```
Comment (CsokiHUN) [inline [esx]/es_extended/server/updater.lua:56]:
```suggestion
    function getRepoInformations()
        local p = promise.new()

        PerformHttpRequest("https://api.github.com/repos/esx-framework/esx-legacy/releases/latest", function(err, response, headers)
            local repoVersion, repoURL, repoBody = curVersion, "https://github.com/esx-framework/esx-legacy", nil
            if err ~= 200 then
              return p:resolve({repoVersion, repoUrl})
            end

            repoVersion = data.tag_name
            repoURL = data.html_url
            p:resolve({repoVersion, repoUrl, data.body})
        end, "GET")

        return table.unpack(Citizen.Await(p))
    end
    ```

File: [esx]/es_extended/server/updater.lua [hunk 2]
```diff
@@ -0,0 +1,57 @@
...
+            Wait(3600000)
+        end
+    end)
+
+    function checkVersion(err, responseText, headers)
+        local repoVersion, repoURL, repoBody = getRepoInformations()
+
+        CreateThread(function()
+            if curVersion ~= repoVersion then
+                Wait(4000)
```
Comment (CsokiHUN) [inline [esx]/es_extended/server/updater.lua:23]: why wait 4 sec?

CsokiHUN (changes requested):

[esx]/es_extended/server/updater.lua
Comment (CsokiHUN) [reply [esx]/es_extended/server/updater.lua:56]: i think promises better than repeat

PR #551, opened 2022-10-09
Create es.lua

nemaDev:


Benzo00:
esx_garage is under going a full rewrite so will not accept any prs for it

nemaDev:
> esx_garage is under going a full rewrite so will not accept any prs for it

ahh okok no problem :)

PR #552, opened 2022-10-09
Create ui_es.html

nemaDev:


Benzo00:
esx_garage is under going a full rewrite so will not accept any prs for it

Issue #343, opened 2022-07-11
PR Guidelines 


Benzo00:
**For those that are trying to contribute to ESX.**

https://www.conventionalcommits.org/en/v1.0.0/

Please follow this, as your PRs will not be accepted nor reviewed until your PR titles are following this properly.

**1.** refactor()
Refactoring is when you change the code, without changing functionality. So minor fixes where the code does the same thing in the end.

**2.** feat()
feat is for all new code.

**3.** fix()
fix is just for fixing mistype, or broken code

**4.** feat!()
all new code, that introduces breaking changes

**5.** fix!()
fix broken code, but introduce breaking changes

```
refactor(esx_lscustom/client/main.lua): fixed thread
feat(esx_lscustom/client/main.lua): added new thread for detecting vehicle type
fix(esx_sit/client/list.lua): fixed mistyped variable
refactor(esx_legacy): replaced all instances of table.insert with table[#table+1]
```

Kenshiin13:
Replaced by [new contribution guidelines](https://github.com/esx-framework/esx_core/blob/main/CONTRIBUTING.md).

PR #557, opened 2022-10-10
refactor(esx_society) add extra translations for other options

nemaDev:


Gittified:
Please take a look at #343

PR #561, opened 2022-10-11
Revert "refactor(esx_society/client/main.lua): now using table.insert()"

hyprell:
Reverts esx-framework/esx-legacy#559, elements[#elements + 1] is faster than table.insert, why would you need to use table.insert if there's faster method?


lexinor:
Didn't know about that. Was just thinking about the code readability

PR #563, opened 2022-10-11
refactor(es_extended/server/classes/player.lua): xPlayer.canCarryItem - Error handling - Item not found

OMikkel:
Recently came across a guy in my community who tried to use this function, but he got this error. 
Error Image

Obviously, this is not that helpful and I would presume that it is not the first time someone passed an unknown item into this function and got this error.

> Added a check to display a more precise error message


Gittified (approved):
Looks good to me

JackDUpModZ (approved):

PR #565, opened 2022-10-11
fix(esx_ambulancejob/client/vehicle.lua): fixed vehicle storage bug

janbiehl:
A vehicle from ambulancejob could be spawned from garage for a unlimited amount. The UI did always tell that the vehicle was stored, no matter what the entry in the Database was stating.

Now after spawning the vehicle, the state is correctly shown in the UI. Multiple spawns are not possible anymore.

### DIFF
--- a/[esx_addons]/esx_ambulancejob/client/vehicle.lua
+++ b/[esx_addons]/esx_ambulancejob/client/vehicle.lua
@@ -50,7 +50,7 @@ function OpenVehicleSpawnerMenu(type, hospital, part, partNum)
 							local vehicleName = GetLabelText(GetDisplayNameFromVehicleModel(props.model))
 							local label = ('%s - <span style="color:darkgoldenrod;">%s</span>: '):format(vehicleName, props.plate)
 
-							if v.stored then
+							if v.stored == 1 then
 								label = label .. ('<span style="color:green;">%s</span>'):format(TranslateCap('garage_stored'))
 							else
 								label = label .. ('<span style="color:darkred;">%s</span>'):format(TranslateCap('garage_notstored'))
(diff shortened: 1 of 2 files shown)


iSentrie:
it breaks for me everything

janbiehl:
> it breaks for me everything

This is crazy, for us it is the only way how we get things to work like expected. In which way it breaks things on your side? What do you see? What do you expect?

iSentrie:
it shows that all vehicles are taken out and i cannot get any. :D

janbiehl:
> it shows that all vehicles are taken out and i cannot get any. :D

For me this is the expected behaviour, a car that is not properly stored in the garage will be shown as impounded. You then have to pay the impound at a towing station. You will get your vehicle back, and you are able to store it back in the garage. 

Without my changes you can take one and the same car out of the garage over and over again. So everybody could drive the same car with the same number plate. This cant be the desired behaviour of this.

iSentrie:
> > it shows that all vehicles are taken out and i cannot get any. :D
> 
> For me this is the expected behaviour, a car that is not properly stored in the garage will be shown as impounded. You then have to pay the impound at a towing station. You will get your vehicle back, and you are able to store it back in the garage.
> 
> Without my changes you can take one and the same car out of the garage over and over again. So everybody could drive the same car with the same number plate. This cant be the desired behaviour of this.

thats how it works without any changes for me, with your changes i cant get a vehicle :D

janbiehl:
> > > it shows that all vehicles are taken out and i cannot get any. :D
> > 
> > 
> > For me this is the expected behaviour, a car that is not properly stored in the garage will be shown as impounded. You then have to pay the impound at a towing station. You will get your vehicle back, and you are able to store it back in the garage.
> > Without my changes you can take one and the same car out of the garage over and over again. So everybody could drive the same car with the same number plate. This cant be the desired behaviour of this.
> 
> thats how it works without any changes for me, with your changes i cant get a vehicle :D thats actually super weird

weird, i think i will close this. As i won't break things for others, maybe there is something strange on our side.

PR #577, opened 2022-10-18
remove nil parameter

kdxr:
maxDistance not pass to this function


MrBadfantomas (approved):

AstroNext:
This function doesn't need maxDistance in parameters.

AstroNext (approved):
right code

PR #578, opened 2022-10-18
fix(esx_banking/client/main.lua): fixed ox_target & fix(esx_multicharacter/client/main.lua): fixed can't rotate camera when making character

ayobpro:


Benzo00:
Ox target is now removed completely

PR #580, opened 2022-10-18
fix(esx_banking): remove ox_target from code

roadtosixx:


coresync2k:
explain why u are removing this? :D no sense @ all if you dont want to use this set it to false?

roadtosixx:
I am in the ESX team, and from above came the request ox_target should be removed.

Benzo00:
> 

No, we have decided either we have target support in all scripts or none, we may or may not have planned to do our own thing :)

PR #581, opened 2022-10-18
Update vehicle.lua

NoNameCfg:
Save cars on sql.


Benzo00:
Please if your going to make a pr then use the translation system and add translation to all locales, once that is done I will review it and either accept it close it

Benzo00:
Open a new one with it correctly made

PR #586, opened 2022-10-22
JSON indentation

katotekii:


CsokiHUN:
Why? This will only increase the file size.

PR #593, opened 2022-10-26
refactor(es_extended): improve readabilty

tomiichx:
- Main changes: /client/functions.lua -> 80-110


[esx]/es_extended/client/functions.lua
```diff
@@ -22,7 +22,7 @@ ESX.Scaleform.Utils = {}
 ESX.Streaming = {}
 
 function ESX.SetTimeout(msec, cb)
-    table.insert(Core.TimeoutCallbacks, {
+    ESX.TableInsert(Core.TimeoutCallbacks, {
```
Comment (Benzo00) [inline [esx]/es_extended/client/functions.lua:25]: ESX.TableInsert?

Comment (Benzo00) [reply [esx]/es_extended/client/functions.lua:25]: oh didnt see that u made a function for it 🤔

Comment (tomiichx) [reply [esx]/es_extended/client/functions.lua:25]: Yes 🙂

[esx]/es_extended/common/functions.lua
```diff
@@ -1,100 +1,105 @@
...
+                if v2.name == weaponComponent then
+                    return v2
+                end
+            end
+        end
+    end
+end
+
+function ESX.TableInsert(table, value)
+    table[#table + 1] = value
```
Comment (Mycroft-Studios) [inline [esx]/es_extended/common/functions.lua:69]: Pretty sure this completely removes the performance benefit lol

Comment (Mycroft-Studios) [reply [esx]/es_extended/common/functions.lua:69]: also, ESX.Table.Insert, would be correct naming, since ESX has a namespace for tables :)

Comment (tomiichx) [reply [esx]/es_extended/common/functions.lua:69]: > Pretty sure this completely removes the performance benefit lol

Indexing the last key and inserting +1 to it as well as it's value would be indeed quicker 🙂

Comment (tomiichx) [reply [esx]/es_extended/common/functions.lua:69]: > also, ESX.Table.Insert, would be correct naming, since ESX has a namespace for tables :)

This could be easily fixed...

Comment (Mycroft-Studios) [reply [esx]/es_extended/common/functions.lua:69]: > > Pretty sure this completely removes the performance benefit lol
> 
> Indexing the last key and inserting +1 to it as well as it's value would be indeed quicker 🙂
 
then why not just use table.insert? it uses the current method because its faster, not because its nicer :)

Comment (tomiichx) [reply [esx]/es_extended/common/functions.lua:69]: I never mentioned that it was prettier, but this is definitely faster. 👍🏼

Comment (tomiichx) [reply [esx]/es_extended/common/functions.lua:69]: You did the same thing here [(977dba4)](https://github.com/esx-framework/esx-legacy/pull/618/commits/977dba4ee3532a404448825fb226a573bc2e26c8), except my commit was meant to make it a global function.

Also, if `ESX.Table.Insert` is the correct name, then `ESX.DumpTable` would also have to be renamed to `ESX.Table.Dump` since we're working with tables/arrays here.

Comment (Mycroft-Studios) [reply [esx]/es_extended/common/functions.lua:69]: > Also, if ESX.Table.Insert is the correct name, then ESX.DumpTable would also have to be renamed to ESX.Table.Dump since we're working with tables/arrays here.

it should be, but that was added 4 years ago, so we cant change it lol

PR #596, opened 2022-10-28
ESX.RegisterInput instead of IsControlPressed

AstroNext:
Remove infinite loop,
Reduce resource usage when menu is open to 0.0
Prevent error and bug when namespace resource restart


iSentrie:
you removed `es_extended/config.lua`

iSentrie (changes requested):
you cant remove `[esx]/es_extended/config.lua`

PR #601, opened 2022-10-29
refactor(esx_menu_default/client/main.lua): Change isControlePress to ESX.RegisterInput

AstroNext:
Reduce usage to 0.0
faster key press then isControlePress
Close all menu's when namespace or resource restarted ( prevent stocked menu )

### DIFF
--- a/[esx]/esx_menu_default/client/main.lua
+++ b/[esx]/esx_menu_default/client/main.lua
@@ -1,7 +1,8 @@
-local GUI, MenuType, OpenedMenus = {}, 'default', 0
+local GUI, MenuType, OpenedMenus, CurrentNameSpace = {}, 'default', 0, nil
 GUI.Time = 0
 
 local function openMenu(namespace, name, data)
+    CurrentNameSpace = namespace
     OpenedMenus += 1
     SendNUIMessage({
         action = 'openMenu',
@@ -12,6 +13,7 @@ local function openMenu(namespace, name, data)
 end
 
 local function closeMenu(namespace, name)
+    CurrentNameSpace = namespace
     OpenedMenus -= 1
     SendNUIMessage({
         action = 'closeMenu',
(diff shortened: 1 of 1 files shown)


Tinky124:
Good job, nice optimization  @AstroNext

[esx]/esx_menu_default/client/main.lua
```diff
@@ -12,6 +13,7 @@ local function openMenu(namespace, name, data)
 end
 
 local function closeMenu(namespace, name)
+    CurrentNameSpace = namespace
```
Comment (Gittified) [inline [esx]/esx_menu_default/client/main.lua:16]: Do you want to set this to **nil** or is it supposed to have the previous namespace?

[esx]/esx_menu_default/client/main.lua
```diff
@@ -20,6 +22,14 @@ local function closeMenu(namespace, name)
     })
 end
 
+AddEventHandler('onResourceStop', function(resource)
+    if GetCurrentResourceName() == resource and OpenedMenus > 0  then
+        ESX.UI.Menu.CloseAll()
+    elseif CurrentNameSpace ~= nil and CurrentNameSpace == resource and OpenedMenus > 0 then
```
Comment (Gittified) [inline [esx]/esx_menu_default/client/main.lua:28]: If CurrentNameSpace does not equal resource it will also automatically mean that it is not nil, so the first check can be removed.

Gittified (changes requested):
Looking pretty good, please take a look at the comments.

[esx]/esx_menu_default/client/main.lua
Comment (AstroNext) [reply [esx]/esx_menu_default/client/main.lua:28]: Most resources use **GetCurrentResourceName()** for name space.
I will check this error handler with this [native](https://docs.fivem.net/natives/?_0x4D52FE5B)

[esx]/esx_menu_default/client/main.lua
Comment (AstroNext) [reply [esx]/esx_menu_default/client/main.lua:16]: I supposed to have the previous namespace.
That's on validate for menu submit.
but now can be **nil**

[esx]/esx_menu_default/client/main.lua
Comment (AstroNext) [reply [esx]/esx_menu_default/client/main.lua:28]: I test native that's return me es_extended.

Gellipapa (approved):
@AstroNext I tested and works, really good job. 
@Gittified What do you think about this?

Taavi-AU:
Unsure if this is needed as esx menu default will be removed soon.

TheFantomas (approved):

Gellipapa:
@AstroNext @Gittified I re test and works perfectly thank you for help.

PR #603, opened 2022-10-30
fix(esx_property): wrong datastore usage

FAr18:
Currently, when the resource starts, it creates redundant rows in `datastore_data`.
The datastore `property` should be non-shared, and we don't need to insert data into `datasotre_data`.


AstroNext (approved):

Mycroft-Studios:
it should be shared, since it has no owner

FAr18:
i only found it for the player's wardrobe, or does it have other uses?
(not even any way to trigger `esx_datastore:getSharedDataStore`)

and it does insert duplicate and useless rows inside my database

Mycroft-Studios:
> or does it have other uses?
nope, the way it works is shit, but it works that way to ensure peoples outfits transfer to v2

FAr18:
got it, thanks for the reply

PR #604, opened 2022-10-30
refactor(esx_ambulancejob): Fixed syncing player

Tinky124:


iSentrie:
Still poor optimization, ambulancejob should be reworked and death animation played instead of ped being actually dead.

iSentrie:
> @iSentrie Since this community is toxic i dont care anymore for esx, and i dont want to use it anymore qbus is better.

there was people who came from qb saying the same, so take care

PR #608, opened 2022-10-31
ESX Basicneeds Config Rework

Kuuzoo:
Just a simple rework of the Config that you have now only one Table where you can define a type, prop etc.

### DIFF
--- a/[esx_addons]/esx_basicneeds/client/main.lua
+++ b/[esx_addons]/esx_basicneeds/client/main.lua
@@ -91,7 +91,6 @@ AddEventHandler('esx_basicneeds:onUse', function(type, prop_name)
 			local boneIndex = GetPedBoneIndex(playerPed, 18905)
 			AttachEntityToEntity(prop, playerPed, boneIndex, 0.12, 0.028, 0.001, 10.0, 175.0, 0.0, true, true, false, true, 1, true)
 
-			print(json.encode(anim))
 			ESX.Streaming.RequestAnimDict(anim.dict, function()
 				TaskPlayAnim(playerPed, anim.dict, anim.name, anim.settings[1], anim.settings[2], anim.settings[3], anim.settings[4], anim.settings[5], anim.settings[6], anim.settings[7], anim.settings[8])
 				RemoveAnimDict(anim.dict)
--- a/[esx_addons]/esx_basicneeds/client/main.lua
+++ b/[esx_addons]/esx_basicneeds/client/main.lua
@@ -6,7 +6,6 @@ AddEventHandler('esx_basicneeds:resetStatus', function()
 	TriggerEvent('esx_status:set', 'thirst', 500000)
 end)
 
-
 RegisterNetEvent('esx_basicneeds:healPlayer')
 AddEventHandler('esx_basicneeds:healPlayer', function()
 	-- restore hunger & thirst
@@ -27,11 +26,10 @@ AddEventHandler('esx:onPlayerSpawn', function(spawn)
 		TriggerEvent('esx_basicneeds:resetStatus')
 	end
 
-	IsDead = false
+	IsDead = false	
 end)
 
 AddEventHandler('esx_status:loaded', function(status)
-
 	TriggerEvent('esx_status:registerStatus', 'hunger', 1000000, '#CFAD0F', function(status)
 		return Config.Visible
 	end, function(status)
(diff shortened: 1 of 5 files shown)


[esx_addons]/esx_basicneeds/client/main.lua
```diff
@@ -114,16 +91,16 @@ AddEventHandler('esx_basicneeds:onDrink', function(prop_name)
 			local boneIndex = GetPedBoneIndex(playerPed, 18905)
 			AttachEntityToEntity(prop, playerPed, boneIndex, 0.12, 0.028, 0.001, 10.0, 175.0, 0.0, true, true, false, true, 1, true)
 
-			ESX.Streaming.RequestAnimDict('mp_player_intdrink', function()
-				TaskPlayAnim(playerPed, 'mp_player_intdrink', 'loop_bottle', 1.0, -1.0, 2000, 0, 1, true, true, true)
-				RemoveAnimDict('mp_player_intdrink')
+			print(json.encode(anim))
```
Comment (CsokiHUN) [inline [esx_addons]/esx_basicneeds/client/main.lua:94]: remove plz

CsokiHUN (changes requested):

CsokiHUN (approved):

[esx_addons]/esx_basicneeds/client/main.lua
Comment (Kuuzoo) [reply [esx_addons]/esx_basicneeds/client/main.lua]: > remove plz

I did remove it. Thank you for telling me to!

TheFantomas (approved):

[esx_addons]/esx_basicneeds/client/main.lua
```diff
@@ -74,38 +71,18 @@ AddEventHandler('esx_basicneeds:isEating', function(cb)
...
-			end)
-		end)
-
-	end
-end)
-
-RegisterNetEvent('esx_basicneeds:onDrink')
-AddEventHandler('esx_basicneeds:onDrink', function(prop_name)
+RegisterNetEvent('esx_basicneeds:onUse')
+AddEventHandler('esx_basicneeds:onUse', function(type, prop_name)
```
Comment (Gellipapa) [inline [esx_addons]/esx_basicneeds/client/main.lua:75]: There should be some way to make the code backwards compatible. If somebody uses this esx_basicneeds:onEat event in another script, the code should not break. 

Do you think there are people who use this event in other scripts? 

A possible solution could be to throw some message that it has changed and make the changes.

What do you think about these?

[esx_addons]/esx_basicneeds/server/main.lua
```diff
@@ -1,34 +1,28 @@
...
+			if v.type == "food" then
+				TriggerClientEvent("esx_status:add", source, "hunger", v.status)
+				TriggerClientEvent('esx_basicneeds:onUse', source, v.type)
+				xPlayer.showNotification(TranslateCap('used_food', ESX.GetItemLabel(k)))
+			elseif v.type == "drink" then
+				TriggerClientEvent("esx_status:add", source, "thirst", v.status)
+				TriggerClientEvent('esx_basicneeds:onUse', source, v.type)
+				xPlayer.showNotification(TranslateCap('used_drink', ESX.GetItemLabel(k)))
+			else
+				print(string.format('^1[ERROR]^0 %s has no correct type defined.', k))
```
Comment (Gellipapa) [inline [esx_addons]/esx_basicneeds/server/main.lua:17]: If the v.type is nil will it not register an item? Think about how this could be solved. It might be enough to throw an deprecated error message.

@Kuuzoo Any ideas for you?

[esx_addons]/esx_basicneeds/client/main.lua
Comment (Gellipapa) [reply [esx_addons]/esx_basicneeds/client/main.lua:75]: I think this is now acceptable. Thanks for the correction.

[esx_addons]/esx_basicneeds/server/main.lua
Comment (Kuuzoo) [reply [esx_addons]/esx_basicneeds/server/main.lua:17]: The Item is getting registered over the "k" variable not over v.type, but also if v.type is not defined it's going with a error print.
I could check if k is a string, but k is getting defined everytime by lua. So if it's not a string it's the index of the table.

So the print would only make sense if k is not a string.

Hope you understood what I meant.

Comment (Gellipapa) [reply [esx_addons]/esx_basicneeds/server/main.lua:17]: Sorry, I said it in a stupid way, so the item will be registered but I thought it's better this way. If you update it, it will use the new Config.Items anyway, no problem. 

No further changes are needed. 
I  merge tomorrow. 
Thank you very much for the pull request.

Comment (Kuuzoo) [reply [esx_addons]/esx_basicneeds/server/main.lua:17]: Thank you for all the tips, your time and just everything ❤️.

Have a great night!
Best regards, kuuzoo!

Benzo00:
tested it and works, Rav3n95 also approved, force pushed by myself since github being weird and didnt let me resolve conflicts, so will close this thanks for the pr Kuuzoo :)

PR #611, opened 2022-11-01
refactor(esx_status): no more need server classes

AstroNext:


Mycroft-Studios (changes requested):
WTF, u just removed the core of status xD

FilipeCuco:
> WTF, u just removed the core of status xD

it is not even loading fxmanifest.lua

Gellipapa:
Hi! @AstroNext What is current status for pull request?  I checked files and i find server/client side CreateStatus functions but fxmanifest don't include these and interesting Config files load only server side, but CreateStatus only use client side. 

Please take a look and we will clear things up :)

AstroNext:
> 

I check codes again,
My experience in this resource tell me that need more edit ( usage on each status or tick when call status function)
this server class is exist in old status resource but not used in server side .

Gellipapa:
@AstroNext  Yes, I saw that the server page is not in use but the config page is. Can you see if you delete the fxmanifest and fix the status? I'm also thinking of older ESX. Backwards compatibility is important.

Gellipapa:
@AstroNext Hi! What's up this refactor? Did you manage to fix it?

TheFantomas:
> @AstroNext Hi! What's up this refactor? Did you manage to fix it?

It looks like it won't fix anything :D That's why I'd probably close it so it won't be useless here :D On the other hand, what is backward compatibility for? However, this part was not started before :D After all, no one could use it without some modifications

Gellipapa:
Yes @TheFantomas. I close if you refactor or rewrite status delete unused files :)

PR #616, opened 2022-11-02
fix(esx_policejob): Fix protection for police job

Tinky124:


Tinky124:
Since this community is toxic i dont care anymore for esx, and i dont want to use it anymore qbus is better.

PR #619, opened 2022-11-03
feat(esx_lscustom): headlights colors

JUX84:
Hi,

This PR replaces the xenon menu and adds colors, with the same price.
modXenon needs to be enabled for colors to work so i enabled automatically and disable it when going back to default.
Xenon default color is "neon".

Regards


TrymTube (changes requested):
make use of the Translation please and add it into english locale

`TranslateCap()`

Gellipapa (approved):
Tested and working.

PR #620, opened 2022-11-03
fix(es_extended): value reset if receiving "false" 

JUX84:
Hi,

Sometimes to reset a vehicle mod, esx_lscustom sends the value "false" in vehicle props.
But "nil" and "false" are ignored by esx.
This change makes it so that only "nil" is ignored but false is sent so we can reset any value.

Regards

### DIFF
--- a/[esx]/es_extended/client/functions.lua
+++ b/[esx]/es_extended/client/functions.lua
@@ -832,62 +832,62 @@ function ESX.Game.SetVehicleProperties(vehicle, props)
         local pearlescentColor, wheelColor = GetVehicleExtraColours(vehicle)
         SetVehicleModKit(vehicle, 0)
 
-        if props.plate then
+        if props.plate ~= nil then
             SetVehicleNumberPlateText(vehicle, props.plate)
         end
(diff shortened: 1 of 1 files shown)


Mycroft-Studios:
this would make no differance? since false and nil have the save value here?

JUX84:
let's say you want to reset the xenon headlights, `esx_lscustom` will send the `false` value.

Currently, as you said, `nil` and `false` are treated the same way when checking in the "if" condition.

The goal of the change is to skip the "if" condition only if the value is `nil`.
If it's `false` we still want to enter in the "if" and run the native, which will reset the headlights in my example.

So yeah, nil and false are treated the same, but when we send `false`, we still want the native to be ran. On the contrary, if the value is `nil` we just skip the native like it's doing currently.

TrymTube (approved):
this looks good and is reasonable since it only checks if it is not `nil` but `0` still passes through

CsokiHUN (approved):

Gellipapa (approved):
Thanks for handle nil value check. I think it's good.

PR #621, opened 2022-11-03
fix(esx_lscustom): default wheels

JUX84:
Hi,

this fixes the default value for wheels.
Currently, if you change to custom wheel type you can never go back to the car default wheels.

Regards

### DIFF
--- a/[esx_addons]/esx_lscustom/client/main.lua
+++ b/[esx_addons]/esx_lscustom/client/main.lua
@@ -237,6 +237,13 @@ function GetAction(data)
                         modType = k,
                         modNum = false
                     })
+                elseif v.modType == 23 then
+                    table.insert(elements, {
+                        label = " " .. TranslateCap('by_default'),
+                        modType = "modFrontWheels",
+                        modNum = -1,
+                        wheelType = -1
+                    })
                 else
                     table.insert(elements, {
                         label = " " .. TranslateCap('by_default'),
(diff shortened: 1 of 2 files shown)


iSentrie:
```lua
price = math.floor(vehiclePrice * data.current.price / 100)
```

`data.current.price` will return a nil

iSentrie (changes requested):
you need to set a price, otherwise it will return a nil and fail on calculating the payment

[esx_addons]/esx_lscustom/client/main.lua
```diff
@@ -237,6 +237,13 @@ function GetAction(data)
                         modType = k,
                         modNum = false
                     })
+                elseif v.modType == 23 then
+                    table.insert(elements, {
+                        label = " " .. TranslateCap('by_default'),
+                        modType = "modFrontWheels",
+                        modNum = -1,
+                        wheelType = -1
```
Comment (iSentrie) [inline [esx_addons]/esx_lscustom/client/main.lua:245]: set additional value
`price = 1`
or something

Comment (JUX84) [reply [esx_addons]/esx_lscustom/client/main.lua]: Hi, thanks for the review.

I put the price to 1 like you said, but I'm thinking of maybe setting the default price in the config so anyone can just change it like the other wheels prices, what do you think?

Comment (iSentrie) [reply [esx_addons]/esx_lscustom/client/main.lua]: > 

config is the proper way to do it 👍

Comment (JUX84) [reply [esx_addons]/esx_lscustom/client/main.lua]: Ok it's pushed, i've set it to 2 by default since the other wheels are at 4+ price, thought 50% would be a good value but if you prefer 1 i change it in a sec

iSentrie (approved):

JackDUpModZ (approved):

PR #650, opened 2022-11-04
feat(esx_society) - et translation

walls9919:
Estonian translation for the resource


iSentrie:
you crazy, could have added all of it to 1 pr you made a big mess

PR #651, opened 2022-11-04
Fix (es_extended): player death event

kdxr:
when player drop from falling and die function not active


TrymTube:
awesome ^^

I done this in a friends newest es_extended and the death not registering was fixed ^^

JackDUpModZ (approved):

PR #652, opened 2022-11-05
refactor(esx_service/server/main.lua) Added condition to handle MaxService when set to -1

lexinor:
- I added a condition to handle when we set MaxService to -1. Because I was not able to take a service when it was set to -1


tony-stark-17:
image
This is how esx_policejob does it which is how it should be done in scripts.

lexinor:
> image This is how esx_policejob does it which is how it should be done in scripts.

So I guess when you put -1 it should not work so ?

lexinor:
Any news ?

lexinor:
I'm closing that PR

PR #668, opened 2022-11-10
feat(esx_blips)

TonoBicicleta:
esx blip creator within the game. for esx framework


Benzo00:
There is better way to do this code, and for right now i will deny this, if you have any questions open a ticket on our esx discord and i will let you know what to improve if you want to do that!

PR #678, opened 2022-11-13
ESX.PlayerData.dead is nil here anyways.

xj0le:


AstroNext:
@xj0le 
ESX.PlayerData.dead is checked for this command

https://github.com/esx-framework/esx-legacy/blob/e31442fa248b401bbb75a673dc07a2ea6cec8d1e/%5Besx%5D/esx_identity/server/main.lua#L542-L561

PR #684, opened 2022-11-18
fix for spawning unlimited job vehicles

Taavi-AU:


TrymTube (approved):

bitpredator (approved):

janbiehl:
Weird i created the same PR a few weekd ago. 
Somebody said that this fix breaks things for him, so i am not the only one with that error :)

TrymTube:
> Weird i created the same PR a few weekd ago. Somebody said that this fix breaks things for him, so i am not the only one with that error :)

I have had this problem too I have put the fix in the Discord before I became Supporter nobody gave attention to it and I forgot

Taavi-AU:
Didn't break anything for me

Taavi-AU:
All it does is changes from checking true/false to 1 or 0

janbiehl:
See here #565 .

By the way the same thing is broken for me in the ambulance job. It is exatly the same as with policejob

Taavi-AU:
Yeah I fixed both, check the commits.

PR #691, opened 2022-11-21
Refactor: some es_extended functions

CsokiHUN:


Rav3n95 (approved):

Gellipapa (approved):
I think it's good. @CsokiHUN tested and works.

PR #696, opened 2022-11-24
fix/feat(es_extended): Basic Statebags + Correct vehicle spawning

Mycroft-Studios:


[esx]/es_extended/server/classes/player.lua
```diff
@@ -19,6 +19,12 @@ function CreateExtendedPlayer(playerId, identifier, group, accounts, inventory,
 	if Config.Multichar then self.license = 'license'.. identifier:sub(identifier:find(':'), identifier:len()) else self.license = 'license:'..identifier end
 
 	ExecuteCommand(('add_principal identifier.%s group.%s'):format(self.license, self.group))
+	
+	Player(self.source).state:set("identifier", self.identifier, true)
+	Player(self.source).state:set("license", self.license, true)
+	Player(self.source).state:set("job", self.job, true)
+	Player(self.source).state:set("group", self.group, true)
+	Player(self.source).state:set("name", self.name, true)
```
Comment (CsokiHUN) [inline [esx]/es_extended/server/classes/player.lua:27]: Why do you keep calling the Player() function?
```suggestion
	local stateBag = Player(self.source).state
	stateBag:set("identifier", self.identifier, true)
	stateBag:set("license", self.license, true)
	stateBag:set("job", self.job, true)
	stateBag:set("group", self.group, true)
	stateBag:set("name", self.name, true)
```

Maybe we can make it so that xPlayer.set is rewritten and uses statebag

Or new function xPlayer.setStatebag/xPlayer.setState

Comment (lexinor) [reply [esx]/es_extended/server/classes/player.lua:27]: I think we could add more states like "handcuffed", "dragged", "onduty" and more

Comment (xorzo1) [reply [esx]/es_extended/server/classes/player.lua:27]: > I think we could add more states like "handcuffed", "dragged", "onduty" and more

🫣

Comment (Mycroft-Studios) [reply [esx]/es_extended/server/classes/player.lua:27]: > Or new function xPlayer.setStatebag/xPlayer.setState
then there would be no point in statebags

Comment (CsokiHUN) [reply [esx]/es_extended/server/classes/player.lua:27]: it would also be a statebag, only simpler. And it would be available from xPlayer

Comment (CsokiHUN) [reply [esx]/es_extended/server/classes/player.lua:27]: But it's just an idea.

xorzo1 (approved):

[esx]/esx_textui/TextUI.lua
```diff
@@ -1,16 +1,21 @@
...
----@param typ string
-local function TextUI(message, typ)
+---@param Type string
+local function TextUI(message, type)
+    IsShowing = true
     SendNUIMessage({
         action = 'show',
         message = message and message or 'ESX-TextUI',
-        type = type(typ) == "string" and typ or 'info'
+        type = type == 0 and "info" or type
```
Comment (CsokiHUN) [inline [esx]/esx_textui/TextUI.lua:10]: An earlier error was the type number, which can only be a string. I think it should be the original.
Or put it in so that it can only be a string type.

CsokiHUN (approved):

PR #706, opened 2022-11-26
fix(es_extended): GetVehicleProperties custom xenon color

vrnagy:
The first return parameter of GetVehicleXenonLightsCustomColor is a boolean indicating if the vehicle has custom color.


Rav3n95 (approved):

paolo-paoluccio:
In the game, police vehicles left lying around in the streets, after the server is closed and reopened, are often no longer available in the police garage (in the garage menu you only see the name of the vehicles with the indication "not in garage" next to them "), and they are not even found in the impound points. Any fix? Maybe there is a problem with Xampp ?

PR #708, opened 2022-11-26
fix(esx_banking)

Gellipapa:
fix(esx_banking/html/scripts/app.js): translate undefined fix
feat(esx_banking/server/main.lua): add new logtransaction export in server side


bitpredator (approved):
Great job, solves the problem correctly.

PR #710, opened 2022-11-27
fix and refactor(esx_taxijob)

Gellipapa:
fix(list car in context menu title is undefined)
fix(put stock functionality)
refactor(client and server side add error handlings, check empty tables)
add(new text in locales)

### DIFF
--- a/[esx]/es_extended/client/functions.lua
+++ b/[esx]/es_extended/client/functions.lua
@@ -1087,6 +1087,10 @@ function ESX.Game.Utils.DrawText3D(coords, text, size, font)
 end
 
 function ESX.ShowInventory()
+    if not Config.EnableDefaultInventory then
+        return
+    end
+
     local playerPed = ESX.PlayerData.ped
     local elements = {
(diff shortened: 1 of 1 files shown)


bitpredator (approved):
Everything works fine

Rav3n95 (approved):

TheFantomas:
Looks good to me. The only problem I have with everyone's access is that they don't put the new locales in all files :D You can probably imagine how annoying it is to find these things :D So I would ask them to put locales in everything.

In this format please
['empty_garage'] = 'No vehicle in the garage!', --Not translated

JackDUpModZ:
@Gellipapa Need to fix conflicts please.

Gellipapa:
Reopen if I get there :)

Unavarage:
There is everything in the locale as I tried it in english and almost every language, the thing is that is still remains undefined, so the prolem is not in the locales but in the script inself.

Gellipapa:
> 

Hi! This is issue closed because we rewrite esx_taxijob we do not fixes.

PR #713, opened 2022-11-27
update add an optional restriction to the order registration

Firou91:
Hi,
I noticed that the registration of a new order on the client side using the `RegisterInput` was systematically accessible to all users. I added an option after all arguments to add a restriction as the command originally allows.


Gellipapa:
HI! @Firou91 Thank you for pull request.  Do you think there would be such an option where only a given rank of ace perm could be used? 

It is written in such a way that the callback can handle whether or not the button is enabled or not. 
Do you think this feature would be useful? 
Where would you use it? 

Thanks for your replies.

Firou91:
> 

Hello,
Sorry for the delay in answering,

The only example I can give you for the moment is this one:
A door creation menu, where I need the `_dev` and the `owner` to be able to access it, so I make a restricted `RegisterInput` and add to my permission file the command access to the `_dev` group. 

Now if I understand correctly, you may be looking for a way to automatically add this new permission to the group. But you have to add a permission on a command as well as a group and create the command in a restricted way.

Sincerely.

Gellipapa:
Hi! @Firou91 Please update your branch and check that it works afterwards.

Gellipapa (approved):

Gellipapa:
@Firou91 Thank you for pull request i add this new pr.

Gellipapa:
@Firou91 Please check this. #828

Benzo00:
Restriced is not used on the client

PR #727, opened 2022-12-04
Update setjob command to use lowercase

Cx-Developments:
using lowercase on the job argument to find a job that may have been typed correctly but with a upper case letter at the start eg "Mechanic" would result in no results even though "mechanic" exists.

### DIFF
--- a/[esx]/es_extended/server/commands.lua
+++ b/[esx]/es_extended/server/commands.lua
@@ -7,8 +7,9 @@ end, false, {help = TranslateCap('command_setcoords'), validate = true, argument
 }})
 
 ESX.RegisterCommand('setjob', 'admin', function(xPlayer, args, showError)
-	if ESX.DoesJobExist(args.job, args.grade) then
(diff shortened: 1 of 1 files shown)


Taavi-AU:
could just do

```lua
args.playerId.setJob(string.lower(args.job), args.grade)
```

no need for a Jobs_Lower var imo

Taavi-AU (changes requested):
could just do

```lua
args.playerId.setJob(string.lower(args.job), args.grade)
```

no need for a Jobs_Lower var imo

Rav3n95 (changes requested):
if ESX.DoesJobExist(args.job, args.grade) then
		args.playerId.setJob(args.job, args.grade)
	elseif ESX.DoesJobExist(string.lower(args.job), args.grade) then
		args.playerId.setJob(string.lower(args.job), args.grade)
	else
		showError(TranslateCap('command_setjob_invalid'))
	end

PR #735, opened 2022-12-09
refactor(es_extended): Server sided Coord Saving + fix spawning of invalid vehicles

Mycroft-Studios:
this is a threadless design that allows both xPlayer.coords and xPlayer.getCoords to correctly function, aswell as removing the need for the client to trigger a server event this is once again, a change that **requires** Onesync **infinity**

### DIFF
--- a/[esx]/es_extended/server/classes/player.lua
+++ b/[esx]/es_extended/server/classes/player.lua
@@ -1,3 +1,9 @@
+local SetTimeout = SetTimeout
+local GetPlayerPed = GetPlayerPed
+local DoesEntityExist = DoesEntityExist
+local GetEntityCoords = GetEntityCoords
+local GetEntityHeading = GetEntityHeading
+
 function CreateExtendedPlayer(playerId, identifier, group, accounts, inventory, weight, job, loadout, name, coords)
 	local targetOverrides = Config.PlayerFunctionOverride and Core.PlayerFunctionOverrides[Config.PlayerFunctionOverride] or {}
 	
@@ -31,7 +37,6 @@ function CreateExtendedPlayer(playerId, identifier, group, accounts, inventory, 
 	end
 
 	function self.setCoords(coords)
-		self.updateCoords(coords)
 		local Ped = GetPlayerPed(self.source)
 		local vector = type(coords) == "vector4" and coords or type(coords) == "vector3" and vector4(coords, 0.0) or
 		vec(coords.x, coords.y, coords.z, coords.heading or 0.0)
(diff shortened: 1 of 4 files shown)


[esx]/es_extended/server/classes/player.lua
```diff
@@ -40,10 +45,23 @@ function CreateExtendedPlayer(playerId, identifier, group, accounts, inventory,
...
+					self.coords = {
+						x = coords.x,
+						y = coords.y, 
+						z = coords.z, 
+						heading = heading or 0.0
+					}
+				end
+			end
+			self.updateCoords()
+		end)
```
Comment (wobozkyng) [inline [esx]/es_extended/server/classes/player.lua:64]: this loop continues even though the player had already disconnected, doesn't it?

Comment (Gellipapa) [reply [esx]/es_extended/server/classes/player.lua:64]: Yes, I also think that the timer will not stop.

Comment (Mycroft-Studios) [reply [esx]/es_extended/server/classes/player.lua:64]: > this loop continues even though the player had already disconnected, doesn't it?

no, since its within the xPlayer class, which only exists when the player is connected, once they use the relog function or disconnect, the function will no longer exist, and therefore, not continue.

Comment (Mycroft-Studios) [reply [esx]/es_extended/server/classes/player.lua:64]: if it uses a thread, that would be the case though, which is why it's a threadless design 🙃

Comment (wobozkyng) [reply [esx]/es_extended/server/classes/player.lua:64]: I had tested it using print, and unfortunately it still printing even I had already disconnected.

Comment (Mycroft-Studios) [reply [esx]/es_extended/server/classes/player.lua:64]: that literally isn't possible 👀, it doesn't exist after you disconnect

MahmoudyTech:
image
I don't know what this problem is

Mycroft-Studios:
> image I don't know what this problem is
this PR fixes that

[esx]/es_extended/server/classes/player.lua
Comment (mahanmoulaei) [reply [esx]/es_extended/server/classes/player.lua:64]: I believe SetTimeout creates a thread internally which would lead to having a lot of threads the more players' count increases. I could be wrong tho...

Comment (Mycroft-Studios) [reply [esx]/es_extended/server/classes/player.lua:64]: > I believe SetTimeout creates a thread internally which would lead to having a lot of threads the more players' count increases if using this method...

as far as i can tell from looking internally, that is not the case :)

bitpredator (approved):

MahmoudyTech:
> > صورةلا أعرف ما هي هذه المشكلة
> 
> يعمل هذا العلاقات العامة على إصلاح ذلك ، المشكلة هي أن السيارة التي تنتجها غير موجودة

The cars are already in the files and with the people in the server and their home is normal

[esx]/es_extended/server/classes/player.lua
Comment (mahanmoulaei) [reply [esx]/es_extended/server/classes/player.lua:64]: > > I believe SetTimeout creates a thread internally which would lead to having a lot of threads the more players' count increases if using this method...
> 
> as far as i can tell from looking internally, that is not the case :)

As far as I can tell that is the case. check [here](https://github.com/citizenfx/fivem/blob/master/code/components/citizen-scripting-lua/src/LuaScriptRuntime.cpp#L1032-L1144)

Issue #753, opened 2022-12-17
[Bug] - esx_script - Issue SetVehicleExtra


icarius:
**Describe the bug**
When you spawn a car all extras are on

**From source**
```
    if props.extras ~= nil then
        for extraId, enabled in pairs(props.extras) do
            **SetVehicleExtra(vehicle, tonumber(extraId), enabled and 0 or 0)**
        end
    end
```
https://github.com/esx-framework/esx-legacy/blob/684752c8aede8d528d1806feb9956edbf8850878/%5Besx%5D/es_extended/client/functions.lua#L878

**With correctionr**
With with correction cars spawn without all extras, only needed or claim with handling
```
    if props.extras ~= nil then
        for extraId, enabled in pairs(props.extras) do
            SetVehicleExtra(vehicle, tonumber(extraId), enabled and 0 or **1**)
        end
    end
```

icarius:
@wobozkyng yes exactly thanks to confirm

PR #755, opened 2022-12-18
fix(es_extended/client/functions.lua): fix swapped SetVehicleExtra enabled parameter

wobozkyng:
parameter is [0=on, 1=off]


wobozkyng:
for #753

icarius:
@wobozkyng Thanks for the confirmation

bitpredator (approved):

PR #770, opened 2022-12-23
fix:(boat license missing)

bitpredator:


CsokiHUN:
I think you added the wrong file, you didn't update the current one.

bitpredator:
> I think you added the wrong file, you didn't update the current one.

i will arrange to fix

TheFarFantomas:
> I think you added the wrong file, you didn't update the current one.

Yes :D He add legacy.sql insted of edit [SQL]/legacy.sql :D

PR #771, opened 2022-12-23
fix:(boat license missing)

bitpredator:


bitpredator:
@CsokiHUN , now it should be correct can you give me confirmation.

CsokiHUN (approved):
Looks good, thanks for the fix.

PR #741, opened 2022-12-14
refactor(esx_lscustom) : fix cache sync with client

AstroNext:
refactor(esx_lscustom\client.lua) : Fix net id to sync on quit server
refactor(esx_lscustom\config.lua) : Fix missed price on hellion transmission mode
refactor(esx_lscustom\server.lua) : Fix missed type on playersCount , fix Customs cache update on event `esx_lscustom:refreshOwnedVehicle`

PR #772, opened 2022-12-23
fix(esx_lscustoms): Sync VehicleProperties State to owned vehicles

Mycroft-Studios:


iSentrie:
#741

Mycroft-Studios:
> #741

this fixes a completely different issue, relating to onesync and server-Sided vehicle spawning

iSentrie:
yeah i tried to point to 13:16 lines, i think both commits had different changes.

PR #778, opened 2022-12-28
Fix Translate Saving

johnsoul777:
Fix Translate Saving


TheFantomas (approved):
Yes, the correction was a typo :D That's why I approve this

image

PR #782, opened 2022-12-28
refactor(esx_service/server/main.lua) Added onduty value in Player statebag

lexinor:
Added onduty value in Player statebag


lexinor:
I think the best way of doing it is to include it directly in the job state bag

lexinor:
Closing it to test it directly in the job data

PR #783, opened 2022-12-28
Update fxmanifest.lua

n1kkec:


Gellipapa:
@n1kkec Hi! Why updated esx_banking version number? This is not 1.9.0 this is current version only esx_banking :)

Issue #779, opened 2022-12-28
[Bug] - esx_property -CCTV


johnsoul777:
**Describe the bug**
When you use the CCTV in a house you can not rotate the camera to the right, it only works left, up and down

**To Reproduce**
1. Go to paleto house or any
2. Click on 'CCTV'
3. press D

**Expected behavior**
make the camera rotate to the right

**Screenshots**
![image](https://user-images.githubusercontent.com/72218928/209810311-2a720e2e-7648-4315-845b-dea0e414313d.png)

**Debug Info (please complete the following information):**
 - OS:  Windows
 - FiveM Artifact: 6163
 - ESX Version: 1.9.0

**Additional context**
Nothing

Gellipapa:
@johnsoul777 Thank you for report! @CsokiHUN fixed your problem, i will merge recently.

PR #785, opened 2022-12-28
fix(esx_property): some fixes and refactor

CsokiHUN:
I checked the #779 , but cant reproduce the bug. 
I fixed what was strange


Rav3n95 (approved):

TheFantomas (approved):
It looks fine

Gellipapa (commented):
Can you delete please version number commits? Need only esx_property commits.  If it's easier for you, create another PR and comment here. Thanks.

PR #790, opened 2022-12-29
Removed unnecessary resources from starting

JackDUpModZ:


server.cfg
```diff
@@ -46,14 +39,13 @@ add_ace resource.es_extended command.add_principal allow
...
 ensure oxmysql
+ensure spawnmanager
 
 ## ESX Legacy
 ensure es_extended
 ensure [esx]
 
 ## ESX Addons
 ensure [esx_addons]
-stop basic-gamemode #remove this if you don't want to use multicharacter
```
Rav3n95, line 59: Should just commented out, since lot of server not using multichar.

server.cfg
```diff
@@ -25,18 +25,11 @@ sets sv_projectDesc "The official recipe of the most popular FiveM RP framework,
...
-set onesync on  # Enable OneSync [REQUIRED FOR LEGACY 1.7.5 +]
-set mysql_connection_string "mysql://user:password@localhost/es_extended?waitForConnections=true&charset=utf8mb4"
-set mysql_ui true
+# set onesync on  # Enable OneSync [REQUIRED FOR LEGACY 1.7.5 +]
+set mysql_connection_string "mysql://user:password@localhost/db_name?waitForConnections=true&charset=utf8mb4"
 
 setr esx:locale "en"
 
-## These resources will start by default.
-ensure chat
```
Rav3n95, line 35: Are you sure we don't need that?

Benzo00:
this would interfere with the recipe.

PR #793, opened 2022-12-30
refactor(esx_vehicleshop)

CsokiHUN:


bitpredator (changes requested):
Hi, @CsokiHUN there seems to be some problem with opening the enterprise vault.

JackDUpModZ (approved):

PR #828, opened 2023-01-09
refactor(RegisterInput)

Gellipapa:
- handle nil values
- add console error message with current resource name
- add ace perm

### DIFF
--- a/[esx]/es_extended/client/functions.lua
+++ b/[esx]/es_extended/client/functions.lua
@@ -193,13 +193,24 @@ else
 end
 
 
-ESX.RegisterInput = function(command_name, label, input_group, key, on_press, on_release)
-    RegisterCommand(on_release ~= nil and "+" .. command_name or command_name, on_press)
-    Core.Input[command_name] = on_release ~= nil and ESX.HashString("+" .. command_name) or ESX.HashString(command_name)
-    if on_release then
-        RegisterCommand("-" .. command_name, on_release)
-    end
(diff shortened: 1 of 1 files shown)


Gellipapa:
#713

Gellipapa:
> hey! Fun fact ,restricted doesn't exist on the client! If u read the docs for Register Command https://docs.fivem.net/natives/?_0x5FA79B0F
> 
> You will see setting to restricted does absolutely nothing since ace perms are completely server sided and cannot be touched on the client 🙂
> 
> > The restricted bool is not used on the client side. Permissions can only be checked on the server side, so if you want to limit your command with an ace permission automatically, make it a server command (by registering it in a server script).

Thanks, I will check again.

[esx]/es_extended/client/functions.lua
```diff
@@ -193,13 +193,24 @@ else
...
-    RegisterKeyMapping(on_release ~= nil and "+" .. command_name or command_name, label, input_group, key)
+ESX.RegisterInput = function(command_name, label, input_group, key, on_press, on_release, restricted)
+
+	if not command_name or not label or not input_group or not key or not on_press then
+		print("[^1ERROR^7] You can't register input because missing parameters! Current resource: [^2"..GetInvokingResource().."^7]")
+		return
+	end
+
+	on_release = on_release ~= nil and on_release or false
+	restricted = restricted ~= nil and restricted or false
```
Comment (Benzo00) [inline [esx]/es_extended/client/functions.lua:204]: restricted isnt used client as mycroft said

PR #829, opened 2023-01-09
fix(ex_extended locales.lua) fix nil value if change locales

Gellipapa:


Gellipapa:
Tested maybe two weeks ago, and works.

PR #834, opened 2023-01-13
refactor(esx_service/server/main.lua): Added some new globalstate stuff and fixed stuff

lexinor:
- esx:playerDropped event was not working properly, source was not working, changed it to playerId
- Added GlobalState[name] to make it easier to track onduty counters

Tested on my local server using 1.9

### DIFF
--- a/[esx_addons]/esx_service/server/main.lua
+++ b/[esx_addons]/esx_service/server/main.lua
@@ -17,11 +17,13 @@ RegisterServerEvent('esx_service:activateService')
 AddEventHandler('esx_service:activateService', function(name, max)
 	InService[name] = {}
 	MaxInService[name] = max
+	GlobalState[name] = GetInServiceCount(name)
 end)
 
 RegisterServerEvent('esx_service:disableService')
 AddEventHandler('esx_service:disableService', function(name)
 	InService[name][source] = nil
+	GlobalState[name] = GetInServiceCount(name)
 end)
 
 RegisterServerEvent('esx_service:notifyAllInService')
@@ -35,12 +37,13 @@ end)
 
 ESX.RegisterServerCallback('esx_service:enableService', function(source, cb, name)
 	local inServiceCount = GetInServiceCount(name)
-
+	
 	if inServiceCount >= MaxInService[name] then
 		cb(false, MaxInService[name], inServiceCount)
 	else
 		InService[name][source] = true
-		cb(true, MaxInService[name], inServiceCount)
+		GlobalState[name] = GetInServiceCount(name)
+		cb(true, MaxInService[name], inServiceCount)		
 	end
 end)
 
@@ -75,8 +78,10 @@ end)
 
 AddEventHandler('esx:playerDropped', function(playerId, reason)
 	for k,v in pairs(InService) do
-		if v[source] == true then
-			v[source] = nil
+		if v[playerId] == true then
+			v[playerId] = nil
+			GlobalState[k] = GetInServiceCount(k)
 		end
 	end
 end)
+  


Taavi-AU (approved):

CsokiHUN:
you added the GlobalState table, but it doesn't make sense because you don't use it anywhere..

Gellipapa:
> 

I think he mean state bags.(GlobalState)

CsokiHUN:
it can be, but it's still useless and the statebag name is wrong

xorzo1 (commented):
Unnecessary changes, not used anywhere.

lexinor:
>

lexinor:
> it can be, but it's still useless and the statebag name is wrong

I guess you didn't understand the meaning that change. This is using global Statebag to track number of players of given job that are currently on duty. Instead of always calling server callback, here you just need to do : GlobalState['police'] and it will give you the current number of police agents that are onduty. So of course it's not used here since this is meant to be used in other scripts.

@xorzo1 Same for you, read what I just wrote

CsokiHUN:
> > it can be, but it's still useless and the statebag name is wrong
> 
> I guess you didn't understand the meaning that change. This is using global Statebag to track number of players of given job that are currently on duty. Instead of always calling server callback, here you just need to do : GlobalState['police'] and it will give you the current number of police agents that are onduty. So of course it's not used here since this is meant to be used in other scripts.
> 
> @xorzo1 Same for you, read what I just wrote

I know how the statebag works, I just didn't think of it first, but the name is not good like that, and I don't see much use for it, a simple export would be enough to synchronize it.
The name is not good because someone may already be using the name 'police', so it must be renamed, e.g. 'police_inservice'

If you write an example of its use and why it is better than a simple export, I will reconsider my opinion

lexinor:
> > > it can be, but it's still useless and the statebag name is wrong
> > 
> > 
> > I guess you didn't understand the meaning that change. This is using global Statebag to track number of players of given job that are currently on duty. Instead of always calling server callback, here you just need to do : GlobalState['police'] and it will give you the current number of police agents that are onduty. So of course it's not used here since this is meant to be used in other scripts.
> > @xorzo1 Same for you, read what I just wrote
> 
> I know how the statebag works, I just didn't think of it first, but the name is not good like that, and I don't see much use for it, a simple export would be enough to synchronize it. The name is not good because someone may already be using the name 'police', so it must be renamed, e.g. 'police_inservice'
> 
> If you write an example of its use and why it is better than a simple export, I will reconsider my opinion

Why using exports when you can simply use State bag that is definitely better for usage like this ? 
You could change the name if you want to I don't mind changing it. But here I used 'name' to fit with existing system which needs to be changed anyways but while waiting using globalstate to store on duty employees work perfectly fine and it's not expensive since it's just updated when players are taking service or not or even when they just exit the game.
If you want in the end we can just rename to something like ("%s_counter"):format(name) but as I said, I wanted to match the existing

CsokiHUN:
Statebag is not the best because it synchronizes to the client side, which results in additional network usage. You still haven't given an example of its use in your solution.

lexinor:
> Statebag is not the best because it synchronizes to the client side, which results in additional network usage. You still haven't given an example of its use in your solution.

I literaly just did. But I will create an example since it's seems like you didn't get it. In my policejob, I want to know how many police agents are currently on duty, instead of calling a whole ESX.TriggerServerCallback which when done server-side is not working properly. Here I just get the GlobalState['police'], simple as that.

lexinor:
That commit has been forgotten when you splitted the repo: https://github.com/esx-framework/esx_service/blob/main/server/main.lua

PR #837, opened 2023-01-14
build: remove esx_phone

jbdevic:


FilipeCuco:
do you have a new phone to add to esx?

jbdevic:
for the moment no
the phone will be moved to a different repo but no longer by default because nobody uses it

MoskalykA:
Personally I propose to keep it but try to improve it over time

CsokiHUN (approved):

CsokiHUN:
If this is removed, we will add [npwd](https://github.com/project-error/npwd) support instead

Rav3n95 (approved):

Gellipapa:
> If this is removed, we will add [npwd](https://github.com/project-error/npwd) support instead

+1 but then you have to add the job notification module.

Benzo00 (approved):

PR #840, opened 2023-01-15
Remove outdated hud & Enhancement(esx_context): Option to keep Control under Nui focus

Rav3n95:
Remove outdated hud  from core


[esx_addons]/esx_society/client/main.lua
```diff
@@ -34,31 +34,14 @@ end)
 
 function EnableSocietyMoneyHUDElement()
 	local societyMoneyHUDElementTpl = '<div><img src="' .. base64MoneyIcon .. '" style="width:20px; height:20px; vertical-align:middle;">&nbsp;{{money}}</div>'
```
Comment (CsokiHUN) [inline [esx_addons]/esx_society/client/main.lua:36]:
```suggestion
```

CsokiHUN (changes requested):

CsokiHUN (approved):

[esx]/esx_context/main.lua
```diff
@@ -1,19 +1,47 @@
...
+				DisableControlAction(0, 69, true)
+				DisableControlAction(0, 70, true)
+				DisableControlAction(0, 91, true)
+				DisableControlAction(0, 92, true)
+				DisableControlAction(0, 95, true)
+				DisableControlAction(0, 98, true)
+				Wait(0)
+			end
+		end)
+	end
```
Comment (CsokiHUN) [inline [esx]/esx_context/main.lua:33]:
````suggestion
	SetNuiFocusKeepInput(keepInput)
	if not keepInput then
		return
	end
	CreateThread(function()
		while keepControl do
			DisableControlAction(0, 1, true)
			DisableControlAction(0, 2, true)
			DisableControlAction(0, 4, true)
			DisableControlAction(0, 6, true)
			DisableControlAction(0, 12, true)
			DisableControlAction(0, 13, true)
			DisableControlAction(0, 24, true)
			DisableControlAction(0, 25, true)
			DisableControlAction(0, 66, true)
			DisableControlAction(0, 67, true)
			DisableControlAction(0, 69, true)
			DisableControlAction(0, 70, true)
			DisableControlAction(0, 91, true)
			DisableControlAction(0, 92, true)
			DisableControlAction(0, 95, true)
			DisableControlAction(0, 98, true)
			Wait(0)
		end
	end)

CsokiHUN (changes requested):

[esx]/esx_context/main.lua
Comment (Rav3n95) [reply [esx]/esx_context/main.lua:33]: SetNuiFocusKeepInput(keepInput) should be under the return too

Comment (CsokiHUN) [reply [esx]/esx_context/main.lua:33]: I thought about this because that's how the focus keep input is restored, but maybe it's not like that

bitpredator (approved):

PR #852, opened 2023-01-22
feat(modules): add npwd support

CsokiHUN:


mcNuggets1:
what is npwd?

CsokiHUN:
[https://github.com/project-error/npwd](https://github.com/project-error/npwd)

Benzo00:
> what is npwd?

A phone that will be replacing esx_phone. a very good phone as well

PR #862, opened 2023-01-26
Feat: Dynamic xPlayer functions + Marker Interactions

Mycroft-Studios:
These are just some functions i did for OnlineM, and thought i would PR them in.

having ESX as a marker "host" system will heavily boost performance, since there wont be so many resources having threads that do the exact same thing, every tick, i will be doing PRs to all the addons to add this in.

Example Usage:

xPlayer functions:
```lua
-- define the function
ESX.PlayerFunction("addXP", function(self, amount)
        local CurrentXP = self.get("xp")
        local NewXP = CurrentXP + amount
        self.set("xp", NewXP)
        self.triggerEvent("onlineM:updateXP")
end)

-- call the function
xPlayer.addXP(1000)
```

Marker Interaction:
```lua
local name = "test"
local coords = vector3(0,0,0)
local draw_distance = 10
local text_ui_text = "[E] Interact"
local drawMarker = true
local key = 38
 ESX.CreateMarker(marker_name, coords, draw_distance, text_ui_text, {
           drawMarker = drawMarker,
           key = key,
            scale = {x = 0.5, y = 0.5, z = 0.5}, -- Scale of the marker
            sprite = 21, -- type of the marker
            colour = {r = 200, g = 50, b = 50, a =255} -- R, G, B, A, colour system
        }, function()
            -- on interaction
        end)

-- to remove the marker:
ESX.RemoveMarker(name)
```

### DIFF
--- a/[core]/es_extended/client/main.lua
+++ b/[core]/es_extended/client/main.lua
@@ -736,7 +736,7 @@ CreateThread(function()
 				local dist = #(PlayerCoords - marker.coords)
 				if dist <= marker.distance then
 					Sleep = 0
-					if marker.show then
+					if marker.settings.drawMarker then
 						DrawMarker(marker.settings.sprite, marker.coords, 0, 0, 0, 0.0, 0, 0, marker.settings.scale.x,
 							marker.settings.scale.y, marker.settings.scale.z, marker.settings.colour.r, marker.settings.colour.g,
 							marker.settings.colour.b, marker.settings.colour.a or 255, false, true, 2, false, nil, nil, false)
@@ -748,7 +748,7 @@ CreateThread(function()
 							ESX.TextUI(marker.helpText, "info")
 						end
 						if marker.action then
-							if IsControlJustPressed(0, marker.key) then
+							if IsControlJustPressed(0, marker.settings.key) then
 								local act = pcall(marker.action)
 								if not act then
 									print("[ERROR] an Error occured during interaction on marker " .. marker.name)
--- a/[core]/es_extended/client/main.lua
+++ b/[core]/es_extended/client/main.lua
@@ -768,3 +768,10 @@ CreateThread(function()
 		Wait(Sleep)
 	end
 end)
+
+-- remove markers if the resource has stopped.
+AddEventHandler("onResourceStop", function(resourceName)
+    if Core.Markers[resourceName] then
+        Core.Markers[resourceName] = nil
+    end
+end)
--- a/[core]/es_extended/client/main.lua
+++ b/[core]/es_extended/client/main.lua
@@ -713,3 +713,58 @@ for i=1, #DoNotUse do
 		print("[^1ERROR^7] YOU ARE USING A RESOURCE THAT WILL BREAK ^1ESX^7, PLEASE REMOVE ^5"..DoNotUse[i].."^7")
 	end
 end
+
+-- Marker Interaction System
+local DrawingTextUI = false
+CreateThread(function()
+	while not ESX.PlayerLoaded do
+		Wait(0)
+	end
+	while true do
+		local Sleep = 1000
+		local ShowTextUI = false
+		local PlayerCoords = GetEntityCoords(ESX.PlayerData.ped)
+		if ESX.Table.SizeOf(Core.Markers) < 1 then
+			goto thread_end -- skip the entire thread
+		end
+		for k, v in pairs(Core.Markers) do
+			if #v < 1 then
+				goto loop_end -- skip this index
+			end
+			for i = 1, #v do
+				local marker = v[i]
+				local dist = #(PlayerCoords - marker.coords)
+				if dist <= marker.distance then
+					Sleep = 0
+					if marker.show then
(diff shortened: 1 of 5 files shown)


CsokiHUN:
Good idea, I support it. However, I think the marker could be a class, similar to xPlayer, only xMarker
with functions and events: playerHit, playerLeave etc

[core]/es_extended/client/main.lua
```diff
@@ -713,3 +713,65 @@ for i=1, #DoNotUse do
...
+local DrawingTextUI = false
+CreateThread(function()
+	while not ESX.PlayerLoaded do
+		Wait(0)
+	end
+	while true do
+		local Sleep = 1000
+		local ShowTextUI = false
+		local PlayerCoords = GetEntityCoords(ESX.PlayerData.ped)
+		if ESX.Table.SizeOf(Core.Markers) < 1 then
```
Comment (CsokiHUN) [inline [core]/es_extended/client/main.lua:727]:
```suggestion
    if not next(Core.Markers) then
```
for better performance 😁

[core]/es_extended/client/main.lua
```diff
@@ -713,3 +713,65 @@ for i=1, #DoNotUse do
 		print("[^1ERROR^7] YOU ARE USING A RESOURCE THAT WILL BREAK ^1ESX^7, PLEASE REMOVE ^5"..DoNotUse[i].."^7")
 	end
 end
+
+-- Marker Interaction System
+local DrawingTextUI = false
+CreateThread(function()
+	while not ESX.PlayerLoaded do
+		Wait(0)
+	end
```
Comment (CsokiHUN) [inline [core]/es_extended/client/main.lua:722]: why not use esx:playerLoaded ?? realy?

[core]/es_extended/server/functions.lua
```diff
@@ -387,6 +387,17 @@ function ESX.SetPlayerFunctionOverride(index)
   Config.PlayerFunctionOverride = index
 end
 
+-- Allows dynamic xPlayer functions, like the overrides but in runtime.
+function ESX.PlayerFunction(name, cb)
```
Comment (CsokiHUN) [inline [core]/es_extended/server/functions.lua:391]: You also need to add the function to the newly created xPlayer in the createExtendedPlayer function

Comment (Mycroft-Studios) [reply [core]/es_extended/server/functions.lua:391]: whoops, forgot that, will add that :)

[core]/es_extended/client/main.lua
Comment (Mycroft-Studios) [reply [core]/es_extended/client/main.lua:722]: > Good idea, I support it. However, I think the marker could be a class, similar to xPlayer, only xMarker with functions and events: playerHit, playerLeave etc

more memory usage, and not really needed :)

Comment (CsokiHUN) [reply [core]/es_extended/client/main.lua:722]: > > Good idea, I support it. However, I think the marker could be a class, similar to xPlayer, only xMarker with functions and events: playerHit, playerLeave etc
> 
> more memory usage, and not really needed :)

better syntax, cleaner codes..
I'm looking forward to other people's opinions.

Gellipapa:
@Mycroft-Studios I think a class based solution would be good, but since there are 0 errors handling with this extended function addition I would not allow to use it. Why should everything be linked to xPlayer?

Mycroft-Studios:
Because that is the entire point of xPlayer????
 To store player data????
What is the point of having it if it can't store player data?????

Rav3n95:
Instead of xp system on xPlayer I would do the metadata system since its very similar and can be used for much more things.

xPlayer:meta('add', 'exp', 100) 
xPlayer:meta('get', 'exp')
xPlayer:meta('remove', 'exp' 100)

Markers could be better as it's own class, xMarker.add, remove, exist etc

I will do similar whit blips

Gellipapa:
> Instead of xp system on xPlayer I would do the metadata system since its very similar and can be used for much more things.
> 
> xPlayer:meta('add', 'exp', 100) xPlayer:meta('get', 'exp') xPlayer:meta('remove', 'exp' 100)
> 
> Markers could be better as it's own class, xMarker.add, remove, exist etc
> 
> I will do similar whit blips

Yes meta is cool key, value. But why add custom function to xPlayer? This is private visibility why want to extended?

Rav3n95:
Metadata would save character data, that's why it's should be on xPlayer

[core]/es_extended/client/main.lua
Comment (mahanmoulaei) [reply [core]/es_extended/client/main.lua:722]: I believe esx:playerLoaded would be a better approach especially in moving towards having an actual multicharacter system that gives you the ability to log out of your character and log back in with another one...

Mycroft-Studios:
> Instead of xp system on xPlayer I would do the metadata system since its very similar and can be used for much more things.
> 
> xPlayer:meta('add', 'exp', 100) 
> xPlayer:meta('get', 'exp')
> xPlayer:meta('remove', 'exp' 100)
> 
> Markers could be better as it's own class, xMarker.add, remove, exist etc
> 
> I will do similar whit blips 

I swear , y'all talking about classes like you just learned what they are . They are cool, but they are not for everything and shouldn't be used for everything lol,
There is literally no need to have an entire class ,just for a marker, it's just wasted memory
And what you just described is exactly what this is

TheFantomas:
Hello, after a professional consultation, we agreed that this pull request will not be accepted.

Thanks for the effort, regards
ESX Developer Team

PR #874, opened 2023-02-01
fix(es_extended/client/main.lua): VehicleProperties networkId validation

WirzRexTon:
Here is the solution to avoid having this in the F8: "Warning : [entity] GetNetworkObject: no object by ID ..."

### DIFF
--- a/[core]/es_extended/client/main.lua
+++ b/[core]/es_extended/client/main.lua
@@ -181,21 +181,17 @@ end)
 
 AddStateBagChangeHandler('VehicleProperties', nil, function(bagName, key, value)
 	if value then
+		Wait(0)
+		local NetId = value.NetId
+		local Tries = 0
+		while not NetworkDoesNetworkIdExist(NetId) and Tries < 2000 do
 			Wait(0)
-			local NetId = value.NetId
-			local Vehicle = NetworkGetEntityFromNetworkId(NetId)
-			local Tries = 0
-			while Vehicle == 0 do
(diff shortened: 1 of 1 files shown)


Mycroft-Studios:
Or just ignore it since its just a debug print

TheFantomas:
Yep 1.0 is too much... I have tested on multiple vehicles and it is not good. You also have to test it on planes that jam the wheel in the road.

Marc85BPG:
> Here is the solution to avoid having this in the F8: "Warning : [entity] GetNetworkObject: no object by ID ..."

Why ignore annoying warnings/ prints.
Thx @WirzRexTon

Gellipapa:
Hi! This is because there are no entity checks. 
Why do you think it should only be checked here? 
Why should you go up to 2000? 
Why not 20-50?

[core]/es_extended/client/main.lua
```diff
@@ -181,21 +181,17 @@ end)
...
-				if Tries > 300 then
-					break
-				end
-			end
-			if NetworkGetEntityOwner(Vehicle) == PlayerId() then
-					ESX.Game.SetVehicleProperties(Vehicle, value)
-			end
+			Tries += 1
+		end
+		local Vehicle = NetworkGetEntityFromNetworkId(NetId)
```
Comment (CsokiHUN) [inline [core]/es_extended/client/main.lua:191]: I think we should add an extra check that if the *Tries* value reaches 2000 then the code should not run any more or we will get an error.

TheFantomas:
Fixed in #885

PR #875, opened 2023-02-01
refactor(es_extended/server/commands.lua): Fixed spawning vehicles in the air

WirzRexTon:
Avoid making the car spawn in the air, it is not visually beautiful.

### DIFF
--- a/[core]/es_extended/server/commands.lua
+++ b/[core]/es_extended/server/commands.lua
@@ -58,7 +58,7 @@ ESX.RegisterCommand('car', 'admin', function(xPlayer, args, showError)
 		{name = "Vehicle", value = args.car, inline = true}
 	})
 
-	ESX.OneSync.SpawnVehicle(args.car, playerCoords, playerHeading, upgrades, function(networkId)
+	ESX.OneSync.SpawnVehicle(args.car, vector3(playerCoords.x, playerCoords.y, playerCoords.z-1), playerHeading, upgrades, function(networkId)
 		if networkId then
 			local vehicle = NetworkGetEntityFromNetworkId(networkId)
 			for i = 1, 20 do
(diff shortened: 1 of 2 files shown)


Mycroft-Studios:
This will cause Vehicles to have spasms and/or get stuck in the ground at times, doing Coords.z - 0.8 would be better 🙃

TheFantomas:
Yep 1.0 is too much... I have tested on multiple vehicles and it is not good. You also have to test it on planes that jam the wheel in the road.

CsokiHUN:
i think big vehicles stuck in ground

Naifen-Phoxidas:
Use PlaceObjectOnGroundProperly(entity)

Tinky124:
> Use PlaceObjectOnGroundProperly(entity)
SetVehicleOnGroundProperly(entity)

TheFantomas:
Client side natives :( That would have to be done on the client side and that's "pointless"...

CsokiHUN:
It might be better to do the grounding on the client side, but only for the admin command

thelindat:
> Client side natives :( That would have to be done on the client side and that's "pointless"...

How is that "pointless"? The server has no concept of the ground, so if you want a vehicle to spawn on the ground then _only the client can do so_. It's not pointless for the same reason setting vehicle properties isn't useless.

Set state on the vehicle marking it as "just spawned", have the owner set it on the ground (and whatever else), then set the state to nil.

TheFantomas:
This post will be closed due to inactivity - 20.2.2023

TheFantomas:
Unfortunately closed due to inactivity - 3 Weeks without any response.

In this state, pr cannot be merged. It doesn't solve the problem, rather it breaks the matter even more.

Thanks to everyone who participated in the development of this pr. If someone wants to fix this, I'd be happy if you create a new pr where this will be fixed.

For the entire developer team
Fantík <3

PR #886, opened 2023-02-10
feat(es_extended): callback functions, vehicle type

CsokiHUN:
### New functions:
- ESX.RegisterClientCallback, ESX.TriggerClientCallback
- ESX.GetVehicleType *(client)*

### Old server callback refactored and moved to the modules folder
Server side ESX.GetVehicleType has been optimized

### DIFF
--- a/[core]/es_extended/client/functions.lua
+++ b/[core]/es_extended/client/functions.lua
@@ -2,8 +2,6 @@ ESX = {}
 Core = {}
 ESX.PlayerData = {}
 ESX.PlayerLoaded = false
-Core.CurrentRequestId = 0
-Core.ServerCallbacks = {}
 Core.Input = {}
 ESX.UI = {}
 ESX.UI.Menu = {}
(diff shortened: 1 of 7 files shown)


Mycroft-Studios:
THIS IS NOT A GOOOD IDEA
client call-backs are vulnerable as shit, the Client should **never** have control over data. this will just lead to cheats being able to abuse resources even easier

CsokiHUN:
> THIS IS NOT A GOOOD IDEA client call-backs are vulnerable as shit, the Client should **never** have control over data. this will just lead to cheats being able to abuse resources even easier

This is not for vital data, only for data such as the type of vehicle

Mycroft-Studios:
> > THIS IS NOT A GOOOD IDEA client call-backs are vulnerable as shit, the Client should **never** have control over data. this will just lead to cheats being able to abuse resources even easier
> 
> This is not for vital data, only for data such as the type of vehicle

Users won't use it that way, that's why I locked it down to that one specific function, because the community won't realise how to use it and just use client data and then boom , suddenly all the work on extra security to esx is useless

CsokiHUN:
Don't include a feature because you think people don't understand it. The basic thing for anyone who makes a script is 'never trust client'. 
Other FiveM frameworks have these features, why not create them?
But we'll wait and see what other people say

PR #906, opened 2023-02-20
Player ped fix

fivemasdasd:


CsokiHUN:
#### Thanks for the PR, but we will not merge.
With the 1.9.1 update, the **esx:playerPedChanged** event. Documentation can be found [here](https://documentation.esx-framework.org/legacy/Common/events/playerpedchanged)

PR #912, opened 2023-02-21
feat: Metadata

Rav3n95:


[core]/es_extended/server/main.lua
```diff
@@ -111,8 +111,9 @@ if not Config.Multichar then
 end
 
 function loadESXPlayer(identifier, playerId, isNew)
-  local userData = {accounts = {}, inventory = {}, job = {}, loadout = {}, playerName = GetPlayerName(playerId), weight = 0}
+  local userData = {accounts = {}, inventory = {}, job = {}, loadout = {}, playerName = GetPlayerName(playerId), weight = 0, meta = {}}
 
+  print(json.encode(loadPlayer,{indent=true}))
```
Comment (CsokiHUN) [inline [core]/es_extended/server/main.lua:116]: remove debug print :)

CsokiHUN (changes requested):

CsokiHUN (approved):
looks good

Gellipapa:
Really cool, but a log system would be better than using prints, and if I have metadata then how can I update the value of a given key? So I don't want the whole table just one value of it, if I saw it right it allows to add table as metadata.

CsokiHUN:
> Really cool, but a log system would be better than using prints, and if I have metadata then how can I update the value of a given key? So I don't want the whole table just one value of it, if I saw it right it allows to add table as metadata.

```lua
xPlayer.setMeta('your_key', value)
```

So you are not updating the whole key, only the value for that key.

Gellipapa:
> 

What happens if I add table as value how do I update the value? 
{bread = 200, water = 300} 
i would like the water to be only 400 the correct output would be: {bread = 200, water = 400}.

Gellipapa:
Another feature, I want to retrieve a value from a table, how do I do that? 

key: table_data
value {bread:200, water:300}

I'm only interested in the water value? How can I get it? 

getMeta(table_data, "water") --> output: 300. 

I think such an option could be in getter, of course it would only work for table.

CsokiHUN:
> What happens if I add table as value how do I update the value? table_data = {bread = 200, water = 300} key: table_data value: {bread = 200, water = 300} i would like the water to be only 400 the correct output would be: {bread = 200, water = 400}.

I understand what you are saying, but I don't think we can do it well. Because what if someone uses a non-string key at the table.. 
It's not that big of a problem if the whole table needs to be updated, because we're not going to store a lot of data here.

Rav3n95:
What if the stored value it's not a table?

CsokiHUN:
If we include the option to retrieve the value of the key from the table, we are in the same place as if we had done it ourselves

Gellipapa:
> Then you get the index of the data, update it yourself, then send the updated table back. That's why I posted in Dev branch, if you have a idea how to fix this issue whit out using the getter first. Go for it :D


A possible solution, with an optional parameter, is to use only a table because you can update a not a table value.
For example: setMeta(table_data,'water', 400)

Gellipapa:
> If it is not a table you return the given key value.

Rav3n95:
Show me the code then :D

Gellipapa:
> If we include the option to retrieve the value of the key from the table, we are in the same place as if we had done it ourselves

It is better than always having to retrieve and index everything. Of course there is also an indexing in the background but it is in the background. It is more practical to have the option to return a specific value or the whole table. 

These are just the features and ideas I threw in here :)

Rav3n95:
`xPlayer.setMeta('bikes', {'20', '24', '26', '27.5', '28', '29'})` How about a not indexed table value?

CsokiHUN:
> xPlayer.setMeta('bikes', {'20', '24', '26', '27.5', '28', '29'}) How about a not indexed table value?

fun fact: you store the numbers as string 😂

Gellipapa:
image

This is core logic. Need nil handler etc.

```lua

local function printTable( t )
 
    local printTable_cache = {}
 
    local function sub_printTable( t, indent )
 
        if ( printTable_cache[tostring(t)] ) then
            print( indent .. "*" .. tostring(t) )
        else
            printTable_cache[tostring(t)] = true
            if ( type( t ) == "table" ) then
                for pos,val in pairs( t ) do
                    if ( type(val) == "table" ) then
                        print( indent .. "[" .. pos .. "] => " .. tostring( t ).. " {" )
                        sub_printTable( val, indent .. string.rep( " ", string.len(pos)+8 ) )
                        print( indent .. string.rep( " ", string.len(pos)+6 ) .. "}" )
                    elseif ( type(val) == "string" ) then
                        print( indent .. "[" .. pos .. '] => "' .. val .. '"' )
                    else
                        print( indent .. "[" .. pos .. "] => " .. tostring(val) )
                    end
                end
            else
                print( indent..tostring(t) )
            end
        end
    end
 
    if ( type(t) == "table" ) then
        print( tostring(t) .. " {" )
        sub_printTable( t, "  " )
        print( "}" )
    else
        sub_printTable( t, "  " )
    end
end


local metadata = {
  ["table_data"] = {bread =  200, water =  300}
}

function getMeta(key, searchedKey)
  if type(metadata[key]) == 'table' and searchedKey then
    return metadata[key][searchedKey]
  end
  return metadata[key]
end

function setMeta(key, value, newValue)
  if type(metadata[key]) == 'table' and newValue then
    metadata[key][value] = newValue
    return
  end
  metadata[key] = value
end

print("GET DEFAULT VALUES")
printTable(getMeta("table_data"))
print("SET NEW VALUE ONLY WATER TO 600")
setMeta("table_data", "water", 600)
print("GET ONLY WATER VALUE")
printTable(getMeta("table_data","water"))
print("------------------------------------")
print("DEFAULT SET VERSION WORKS {bread =  200, water =  300}")
setMeta("table_data", {bread =  200, water =  300})
print("GET DEFAULT VALUES")
printTable(getMeta("table_data"))



 ```

**Comment by Gellipapa** [comment]
> 

You can do index, but why do this or you can use default version can set all table -> backwards compatiblity. Sorry guys i should go to work.

**Comment by Gellipapa** [comment]
Next level if i can search in one value, why can't we pass it a table if i only need 5 out of 20 values, same with update, it would be nice to update those 5 values in order. 👍

**Comment by Rav3n95** [comment]
Also play a violin and a guitar in a same time ;)

**Comment by Rav3n95** [comment]
`
RegisterCommand('meta', function(source)
    local xPlayer = ESX.GetPlayerFromId(source)
    print('--------------------------------------------')
    print('Clearing "shopPrices" meta')
    xPlayer.clearMeta('shopPrices')
    print('First meta getter')
    print(json.encode(xPlayer.getMeta(),{indent=true}))
    print('--------------------------------------------')
    print('Setting "shopPrices" meta')
    xPlayer.setMeta('shopPrices', {bread = 200, water = 300})
    print('--------------------------------------------')
    print('Second meta getter')
    print(json.encode(xPlayer.getMeta(),{indent=true}))
    print('--------------------------------------------')
    print('Setting "bread" value on "shopPrices" meta')
    xPlayer.setMeta('shopPrices', 'bread', 250)
    print('Third meta getter -> just getting the bread price')
    print(xPlayer.getMeta('shopPrices', 'bread'))
    print('--------------------------------------------')
    print('Forth meta getter')
    print(json.encode(xPlayer.getMeta(),{indent=true}))
end, false)
`

**Review by Gellipapa — APPROVED** [review]
Nice good job, thank you buddy. 👍

**Review by TheFantomas — APPROVED** [review]

**Comment by gerung** [comment]
cool, its posible to direct save when set meta?

**Comment by Rav3n95** [comment]
> cool, its posible to direct save when set meta?

If you are using the ESX.Saveplayer() function after you set the meta 🤣

**Comment by gerung** [comment]
> > cool, its posible to direct save when set meta?
> 
> If you are using the ESX.Saveplayer() function after you set the meta 🤣

i just set player detail (first lastname, bod, sex, height, etc) as a meta, but i have issue when get the meta, give me nil value
```
GetCharacters = function(source,data,slots)
	local characters = {}
	local license = ESX.GetIdentifier(source)
	local id = Config.Prefix..'%:'..license
	local data = MySQL.query.await('SELECT * FROM users WHERE identifier LIKE ?', {'%'..id..'%'})
	if data then
		for k,v in pairs(data) do
			local job, grade = v.job or 'imigran', tostring(v.job_grade)
			local accounts = json.decode(v.accounts)
			local id = tonumber(string.sub(v.identifier, #Config.Prefix+1, string.find(v.identifier, ':')-1))
			local xPlayer = ESX.GetPlayerFromId(source)
			print(xPlayer.getMeta('detail','firstname'))
		end
	end
	return {characters = characters , slots = slots}
end
```

**Comment by Gellipapa** [comment]
> > > cool, its posible to direct save when set meta?
> > 
> > 
> > If you are using the ESX.Saveplayer() function after you set the meta 🤣
> 
> i just set player detail (first lastname, bod, sex, height, etc) as a meta, but i have issue when get the meta, give me nil value
> 
> ```lua
> GetCharacters = function(source,data,slots)
> 	local characters = {}
> 	local license = ESX.GetIdentifier(source)
> 	local id = Config.Prefix..'%:'..license
> 	local data = MySQL.query.await('SELECT * FROM users WHERE identifier LIKE ?', {'%'..id..'%'})
> 	if data then
> 		for k,v in pairs(data) do
> 			local job, grade = v.job or 'imigran', tostring(v.job_grade)
> 			local accounts = json.decode(v.accounts)
> 			local id = tonumber(string.sub(v.identifier, #Config.Prefix+1, string.find(v.identifier, ':')-1))
> 			local xPlayer = ESX.GetPlayerFromId(source)
> 			print(xPlayer.getMeta('detail','firstname'))
> 		end
> 	end
> 	return {characters = characters , slots = slots}
> end
> ```

Hi! If you print json.encode(xPlayer.getMeta('detail')) what is get?

**Comment by gerung** [comment]
> > > > cool, its posible to direct save when set meta?
> > > 
> > > 
> > > If you are using the ESX.Saveplayer() function after you set the meta 🤣
> > 
> > 
> > i just set player detail (first lastname, bod, sex, height, etc) as a meta, but i have issue when get the meta, give me nil value
> > ```lua
> > GetCharacters = function(source,data,slots)
> > 	local characters = {}
> > 	local license = ESX.GetIdentifier(source)
> > 	local id = Config.Prefix..'%:'..license
> > 	local data = MySQL.query.await('SELECT * FROM users WHERE identifier LIKE ?', {'%'..id..'%'})
> > 	if data then
> > 		for k,v in pairs(data) do
> > 			local job, grade = v.job or 'imigran', tostring(v.job_grade)
> > 			local accounts = json.decode(v.accounts)
> > 			local id = tonumber(string.sub(v.identifier, #Config.Prefix+1, string.find(v.identifier, ':')-1))
> > 			local xPlayer = ESX.GetPlayerFromId(source)
> > 			print(xPlayer.getMeta('detail','firstname'))
> > 		end
> > 	end
> > 	return {characters = characters , slots = slots}
> > end
> > ```
> 
> Hi! If you print json.encode(xPlayer.getMeta('detail')) what is get?

image

**Comment by gerung** [comment]
> > > > cool, its posible to direct save when set meta?
> > > 
> > > 
> > > If you are using the ESX.Saveplayer() function after you set the meta 🤣
> > 
> > 
> > i just set player detail (first lastname, bod, sex, height, etc) as a meta, but i have issue when get the meta, give me nil value
> > ```lua
> > GetCharacters = function(source,data,slots)
> > 	local characters = {}
> > 	local license = ESX.GetIdentifier(source)
> > 	local id = Config.Prefix..'%:'..license
> > 	local data = MySQL.query.await('SELECT * FROM users WHERE identifier LIKE ?', {'%'..id..'%'})
> > 	if data then
> > 		for k,v in pairs(data) do
> > 			local job, grade = v.job or 'imigran', tostring(v.job_grade)
> > 			local accounts = json.decode(v.accounts)
> > 			local id = tonumber(string.sub(v.identifier, #Config.Prefix+1, string.find(v.identifier, ':')-1))
> > 			local xPlayer = ESX.GetPlayerFromId(source)
> > 			print(xPlayer.getMeta('detail','firstname'))
> > 		end
> > 	end
> > 	return {characters = characters , slots = slots}
> > end
> > ```
> 
> Hi! If you print json.encode(xPlayer.getMeta('detail')) what is get?

works, i need to define like this
local meta = xPlayer.getMeta('detail') and meta.firstname, etc.  but i get a new issue, when i try to logout, i get the same issue

**Comment by gerung** [comment]
> > > > cool, its posible to direct save when set meta?
> > > 
> > > 
> > > If you are using the ESX.Saveplayer() function after you set the meta 🤣
> > 
> > 
> > i just set player detail (first lastname, bod, sex, height, etc) as a meta, but i have issue when get the meta, give me nil value
> > ```lua
> > GetCharacters = function(source,data,slots)
> > 	local characters = {}
> > 	local license = ESX.GetIdentifier(source)
> > 	local id = Config.Prefix..'%:'..license
> > 	local data = MySQL.query.await('SELECT * FROM users WHERE identifier LIKE ?', {'%'..id..'%'})
> > 	if data then
> > 		for k,v in pairs(data) do
> > 			local job, grade = v.job or 'imigran', tostring(v.job_grade)
> > 			local accounts = json.decode(v.accounts)
> > 			local id = tonumber(string.sub(v.identifier, #Config.Prefix+1, string.find(v.identifier, ':')-1))
> > 			local xPlayer = ESX.GetPlayerFromId(source)
> > 			print(xPlayer.getMeta('detail','firstname'))
> > 		end
> > 	end
> > 	return {characters = characters , slots = slots}
> > end
> > ```
> 
> Hi! If you print json.encode(xPlayer.getMeta('detail')) what is get?

image
still error

**Comment by Rav3n95** [comment]
> > > > > cool, its posible to direct save when set meta?
> > > > 
> > > > 
> > > > If you are using the ESX.Saveplayer() function after you set the meta 🤣
> > > 
> > > 
> > > i just set player detail (first lastname, bod, sex, height, etc) as a meta, but i have issue when get the meta, give me nil value
> > > ```lua
> > > GetCharacters = function(source,data,slots)
> > > 	local characters = {}
> > > 	local license = ESX.GetIdentifier(source)
> > > 	local id = Config.Prefix..'%:'..license
> > > 	local data = MySQL.query.await('SELECT * FROM users WHERE identifier LIKE ?', {'%'..id..'%'})
> > > 	if data then
> > > 		for k,v in pairs(data) do
> > > 			local job, grade = v.job or 'imigran', tostring(v.job_grade)
> > > 			local accounts = json.decode(v.accounts)
> > > 			local id = tonumber(string.sub(v.identifier, #Config.Prefix+1, string.find(v.identifier, ':')-1))
> > > 			local xPlayer = ESX.GetPlayerFromId(source)
> > > 			print(xPlayer.getMeta('detail','firstname'))
> > > 		end
> > > 	end
> > > 	return {characters = characters , slots = slots}
> > > end
> > > ```
> > 
> > 
> > Hi! If you print json.encode(xPlayer.getMeta('detail')) what is get?
> 
> image still error, fyi: i have string and number value on detail meta

Why don't you use the data from sql? v.meta its there is v.job

**Comment by Rav3n95** [comment]
Also you have a modified server bud, don't be surprise.

**Comment by Gellipapa** [comment]
@gerung It's works perfectly, i think you've got the problem.
image

**Comment by gerung** [comment]
> @gerung It's works perfectly, i think you've got the problem. image

I got an issue when I first spawn when the player already loaded everything fine, but when first spawn xPlayer was not ready to read so I got the issue. I'm using multicharacter and I thing that is the issue

**Comment by Gellipapa** [comment]
> 

I also use multicharacter, I don't understand what first spawn is. So when you see your character you already have xPlayer so you will getMeta as well as other xPlayer function. It won't before since you get xPlayer in the playerLoad event.

**Comment by Gellipapa** [comment]
I assume you mean when you're in the "lobby" or when you're choosing a character. There will be no xPlayer there yet so yes getMeta will not work as well as xPlayer.identifier because xPlayer is nil.

**Comment by Gellipapa** [comment]
If you still don't have this problem, write down a sequence of reproducible steps that we can use to trigger the problem.

**Comment by CsokiHUN** [comment]
For multicharacter, as long as the player is not spawned, only the character is in the selector, xPlayer has no value in my opinion

**Comment by gerung** [comment]
> I assume you mean when you're in the "lobby" or when you're choosing a character. There will be no xPlayer there yet so yes getMeta will not work as well as xPlayer.identifier because xPlayer is nil.
> 
> If this is your problem you will have to write a database query to get the meta data. It's like a column that can be queried by just json.decode the contents of the column to get a usable LUA table that you can index later.

agree, but currently, i don't have any idea how to do this, if you have sample metadata on multicharacter give me some logic, thx

**Comment by gerung** [comment]
> I assume you mean when you're in the "lobby" or when you're choosing a character. There will be no xPlayer there yet so yes getMeta will not work as well as xPlayer.identifier because xPlayer is nil.
> 
> If this is your problem you will have to write a database query to get the meta data. It's like a column that can be queried by just json.decode the contents of the column to get a usable LUA table that you can index later.

solved, thanks

**Comment by gerung** [comment]
@Rav3n95 is there an option to run on the client side?

**Comment by gerung** [comment]
@Rav3n95 is there any option to get meta on client side?

**Comment by Rav3n95** [comment]
If you are using Legacy it's synced into Playerdata

**Comment by gerung** [comment]
> ```lua
> getMeta
> ```

and it's possible to update from client side?

**Comment by ucingoyen** [comment]
i have issue when set the meta, for example i have this meta on db
`{"chardata":{"dateofbirth":"2023-03-14","height":100,"injail":0,"isdead":false,"firstname":"Herman","ishandcuffed":false,"lastname":"Ringgo","ispunishment":0,"sex":"m"}}`
and trying to test with this command
```lua
RegisterCommand('test', function(source)
    local src = source
    local xPlayer = ESX.GetPlayerFromId(src)
    xPlayer.setMeta('chardata', 'ishandcuffed', true)
end, true)
```
but the meta changed like this `{"chardata":{"ishandcuffed":true}}`

hmm

Rav3n95:
Are you using the Dev branch of esx?

ucingoyen:
> Are you using the Dev branch of esx?

no, using the "Main" esx

Rav3n95:
It's not even on the main yet

ucingoyen:
> It's not even on the main yet

i just tried your metadata on main esx

Rav3n95:
You forgot to copy something then, cuz it's already been use on a live server

Gellipapa:
[es_extended_currenty_dev.zip](https://github.com/esx-framework/esx_core/files/10892002/es_extended_currenty_dev.zip)

@ucingoyen Hi! Download this zip and drag and drop and overwrite your es_extended(but make backup save) this is include metadata system.

PR #915, opened 2023-02-23
fix(es_extended/client/functions) Fix infinite loop 

Kenshiin13:
Collision detection does not always yield a result. I didn't do much testing, but it certainly always happens for non-networked addon vehicles when the player is too far away. To be safe, I suggest to just force break out of the loop after a certain amount of tries. Checking every frame seems excessive imo.

### DIFF
--- a/[core]/es_extended/client/functions.lua
+++ b/[core]/es_extended/client/functions.lua
@@ -432,8 +432,13 @@ function ESX.Game.SpawnVehicle(vehicle, coords, heading, cb, networked)
         SetVehRadioStation(vehicle, 'OFF')
 
         RequestCollisionAtCoord(vector.xyz)
+        local tries = 0
         while not HasCollisionLoadedAroundEntity(vehicle) do
-            Wait(0)
+            if tries > 20 then
+                break
+            end
+            tries += 1
+            Wait(50)
         end
 
         if cb then


mcNuggets1:
very good

Kenshiin13:
I understand that this is the wrong usecase, but I believe an uncaught infinite loop is always a bad idea. I am fine with throwing an error and not running the loop at all, but leaving it like this is pointless. It took me a while of debugging to realize why es_extended runs on like 0.13ms in resmon, and this, or rather the scripts that called that function without distance checking, was the reason. https://github.com/Sub-Zero-Interactive/esx_core/commit/eeadbbc6d4d4bb10516c5cf6d27ff2790b344bf1 Is a good compromise tho.

mcNuggets1:
not fixing something because the use case is different should be criminal lmao.

Mycroft-Studios:
> not fixing something because the use case is different should be criminal lmao.

its not a bug if it isnt meant to be used that way..
thats called user error :)

Kenshiin13:
> > not fixing something because the use case is different should be criminal lmao.
> 
> its not a bug if it isnt meant to be used that way.. thats called user error :)

Unhandled user error is a bug.

TheFantomas:
Hello,

We apologize for the rather long delay. Our answer comes this late.

Mycroft is right about this. This error happens if you spawn the car outside the so-called onesync zone. This was fixed in a PR by Mycroft and so it shouldn't happen anymore as there is error handling.

Still, thank you for your willingness to improve the ESX Framework.

With respect and regards
Legacy Management/Leader & Legacy/Infinity Developer
Fantík & TheKuca

PR #918, opened 2023-02-27
Fix: Metadata update to client side

Rav3n95:


gerung:
how to update from client side? it can be done using 'ESX.SetPlayerData`?

PR #921, opened 2023-03-01
feat(es_extended/common/modules/table): Added a better inserting func…

Thekuca:
…tion

Added a better inserting function to use instead of table.insert(table, value)


CsokiHUN:
I think it's useless, the faster the speed would be compared to table.insert, the slower it will work from the ESX object

PR #922, opened 2023-03-01
Feat/Fix(es_extended & esx_multicharacter): Added Validations + bug fixes

Mycroft-Studios:


[core]/es_extended/client/functions.lua
```diff
@@ -406,6 +406,12 @@ function ESX.Game.SpawnVehicle(vehicle, coords, heading, cb, networked)
     local model = type(vehicle) == 'number' and vehicle or joaat(vehicle)
     local vector = type(coords) == "vector3" and coords or vec(coords.x, coords.y, coords.z)
     networked = networked or true
+    local player_coords = GetEntityCoords(ESX.PlayerData.ped)
```
Comment (Gellipapa) [inline [core]/es_extended/client/functions.lua:409]: vector or player_coords or ESX.PlayerData.ped is nil break code

[core]/es_extended/server/main.lua
```diff
@@ -1,6 +1,7 @@
 SetMapName('San Andreas')
 SetGameType('ESX Legacy')
 
+local OnesyncState = GetConvar('onesync', 'off')
```
Comment (Gellipapa) [inline [core]/es_extended/server/main.lua:4]: I think you should name is oneSyncState.

[core]/esx_multicharacter/server/main.lua
```diff
@@ -1,13 +1,18 @@
 if not ESX then
 	error('\n^1Unable to start Multicharacter - you must be using ESX Legacy^0')
 elseif ESX.GetConfig().Multichar then
+	local DatabaseConnected = false
```
Comment (Gellipapa) [inline [core]/esx_multicharacter/server/main.lua:4]: This full logic should not be used here because if there is no database connection then es_extended cannot retrieve the data from the database so it is useless to look elsewhere to see what is happening with the database. This should be tested in es_extended, but even oxmysql handles the sql connection so even there you don't need to just make sure that if there is no database connection then the code doesn't run.

[core]/esx_multicharacter/server/main.lua
```diff
@@ -159,11 +159,12 @@ elseif ESX.GetConfig().Multichar then
 			local count = 0
 
 			for i = 1, #DB_COLUMNS do
-				local v = DB_COLUMNS[i]
+				local column = DB_COLUMNS[i]
+				DB_TABLES[column.TABLE_NAME] = column.COLUMN_NAME
```
Comment (Gellipapa) [inline [core]/esx_multicharacter/server/main.lua:163]: DB_TABLES where use? I don't know what is this.

Gellipapa (commented):
Please read these comments!

[core]/esx_multicharacter/server/main.lua
Comment (Mycroft-Studios) [reply [core]/esx_multicharacter/server/main.lua:163]: > DB_TABLES where use? I don't know what is this.

its used for deleting characters. it loops through, the tables and deletes all the user data from the database. you probably dont recognise it because its been broken like this since the commit where it got added into the main repo, during that transition there was an update that accidently broke it. ive known about it for ages, its just been too simple of a fix for me to put the effort into PRing :)

[core]/es_extended/server/main.lua
Comment (Mycroft-Studios) [reply [core]/es_extended/server/main.lua:4]: 👍🏼

[core]/es_extended/client/functions.lua
Comment (Mycroft-Studios) [reply [core]/es_extended/client/functions.lua:409]: ?

[core]/esx_multicharacter/server/main.lua
Comment (Mycroft-Studios) [reply [core]/esx_multicharacter/server/main.lua:4]: that isnt what this code is for :)

This is for selecting all tables that use owner and/or identifier which is required for multicharacter :)

Comment (Mycroft-Studios) [reply [core]/esx_multicharacter/server/main.lua:4]: its an automated version of what the original from esx_kashacters (what esx_multicharacter is a fork of). in the original you had to put all of them into the config manually.

[core]/es_extended/client/functions.lua
Comment (Gellipapa) [reply [core]/es_extended/client/functions.lua:409]: You calculate distance but if player_coords or vector is nil break code. You should handle nil values.

[core]/esx_multicharacter/server/main.lua
Comment (Gellipapa) [reply [core]/esx_multicharacter/server/main.lua:4]: Hmm interesting but why do we have to search at this level to use the database connection string? If you have the connection you can query the database to see if there is a table or not with a plain inform_ default sql table.

[core]/esx_multicharacter/server/main.lua
Comment (Gellipapa) [reply [core]/esx_multicharacter/server/main.lua:163]: I understand there is no better way to do this? If everything is already linked to everything and we know what table the player data can be in, then delete it with a query?

[core]/es_extended/client/functions.lua
Comment (TheFantomas) [reply [core]/es_extended/client/functions.lua:409]: use camelCase please.

Mycroft-Studios:
Going Todo some follow up commits later today

TheFantomas:
> Going Todo some follow up commits later today

Good :D

[core]/esx_multicharacter/server/main.lua
Comment (Mycroft-Studios) [reply [core]/esx_multicharacter/server/main.lua]: Because normal sql queries with oxmysql act within the database, we need the data from outside of the database, to get table names and there columns.

Comment (Gellipapa) [reply [core]/esx_multicharacter/server/main.lua]: > Because normal sql queries with oxmysql act within the database, we need the data from outside of the database, to get table names and there columns.

It's query get current database but okey  SELECT DATABASE(); No problem. Thanks for describing why need this.

[core]/esx_multicharacter/server/main.lua
Comment (Mycroft-Studios) [reply [core]/esx_multicharacter/server/main.lua:160]: This is the best way to do this, yes. Any other way would require all scripts to have code in for deletion and would end up with next to no adoption from the community, which would make deleting a character useless, since all the character data would still be in the database 🥲

Rav3n95 (approved):

PR #925, opened 2023-03-03
feat(es_extended/common/modules): TableContains

baguscodestudio:
Example usage:
`ESX.Table.TableContains(category, {'plane', 'helicopter'})` 
`ESX.Table.TableContains(category, 'boat')`

Just a small function that might be neat to have

### DIFF
--- a/[core]/es_extended/common/modules/table.lua
+++ b/[core]/es_extended/common/modules/table.lua
@@ -133,6 +133,25 @@ function ESX.Table.Join(t, sep)
 	return str
 end
 
+-- Credits: https://github.com/JonasDev99/qb-garages/blob/b0335d67cb72a6b9ac60f62a87fb3946f5c2f33d/server/main.lua#L5
+function ESX.Table.TableContains(tab, val)
+    if type(val) == "table" then
+        for _, value in ipairs(tab) do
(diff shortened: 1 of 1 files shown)


[core]/es_extended/common/modules/table.lua
```diff
@@ -133,6 +133,25 @@ function ESX.Table.Join(t, sep)
 	return str
 end
 
+-- Credits: https://github.com/JonasDev99/qb-garages/blob/b0335d67cb72a6b9ac60f62a87fb3946f5c2f33d/server/main.lua#L5
+function ESX.Table.TableContains(tab, val)
+    if type(val) == "table" then
+        for _, value in ipairs(tab) do
```
Comment (MoskalykA) [inline [core]/es_extended/common/modules/table.lua:139]: Be careful with the ipairs operation, it does not work in all cases https://stackoverflow.com/questions/55108794/what-is-the-difference-between-pairs-and-ipairs-in-lua

baguscodestudio:
> Alternatively, I think we can use the table key where the value is found as the return value

Would it be better to have it as an option like we can pass another argument to return as value or just boolean?
Example would be:
`ESX.Table.TableContains(category, {'plane', 'helicopter'}, true)`
So third argument accepts boolean as a parameter to return the value or not, if its false then it will just return the boolean.

PR #931, opened 2023-03-08
feat(es_extended/config.lua): Adding disable seat shuff (PART 1)

Thekuca:


CoachYT1:
Why are you making multiple pull requests to add the same functionality when you can add commits to a single pull request?

PR #933, opened 2023-03-11
refactor(sharedObject): Add event back to es_extended.

Flambaa:
This pull requests adds a config option to enable the old getSharedObject event. The amount of people who do not know how to use the the new export is ridiculous and we all know things wont change. Script developers wont change their escrowed code either, meaning people cannot use scripts that they have paid for. 

I believe this will solve a lot of our problems with backwards compatibility as well as minimise the amount of people asking about the error in our discord.

Feel free to roast the idea or put your opinions/thoughts in the comments. Please no smart ass or rude comments, be respectful! :)

### DIFF
--- a/[core]/es_extended/client/common.lua
+++ b/[core]/es_extended/client/common.lua
@@ -6,7 +6,11 @@ if GetResourceState('ox_inventory') ~= 'missing' then
 	Config.OxInventory = true
 end
 
-AddEventHandler("esx:getSharedObject", function()
-	local Invoke = GetInvokingResource()
-	print(("[^1ERROR^7] Resource ^5%s^7 Used the ^5getSharedObject^7 Event, this event ^1no longer exists!^7 Visit https://documentation.esx-framework.org/tutorials/tutorials-esx/sharedevent for how to fix!"):format(Invoke))
+AddEventHandler("esx:getSharedObject", function(cb)
+	if Config.EnableGetSharedObjectEvent then
+		cb(ESX)
+	else
+		local Invoke = GetInvokingResource()
+		print(("[^1ERROR^7] Resource ^5%s^7 Used the ^5getSharedObject^7 Event, this event ^1no longer exists!^7 Visit https://documentation.esx-framework.org/tutorials/tutorials-esx/sharedevent for how to fix! Alternativly you can enable the config option for it!"):format(Invoke))
+	end
 end)
--- a/[core]/es_extended/client/common.lua
+++ b/[core]/es_extended/client/common.lua
@@ -11,6 +11,6 @@ AddEventHandler("esx:getSharedObject", function(cb)
 		cb(ESX)
 	else
 		local Invoke = GetInvokingResource()
-		print(("[^1ERROR^7] Resource ^5%s^7 Used the ^5getSharedObject^7 Event, this event ^1no longer exists!^7 Visit https://documentation.esx-framework.org/tutorials/tutorials-esx/sharedevent for how to fix! Alternativly you can enable the config option for it!"):format(Invoke))
+		print(("[^1ERROR^7] Resource ^5%s^7 Used the ^5getSharedObject^7 Event, This Is NOT Recommended and is very bad for performance! Visit https://documentation.esx-framework.org/tutorials/tutorials-esx/sharedevent For More details"):format(Invoke))
 	end
 end)
--- a/[core]/es_extended/config.lua
+++ b/[core]/es_extended/config.lua
@@ -63,7 +63,7 @@ Config.RemoveHudCommonents = {
 	[22] = false, --HUD_WEAPONS
 }
 
-Config.EnableGetSharedObjectEvent = false -- This is not recommended, Enables old getSharedObject event.
+Config.EnableGetSharedObjectEvent = false -- Warning: Enabling This WILL massively hurt and decrease your servers performance, use this at your own precaution.
 
 Config.SpawnVehMaxUpgrades = true -- admin vehicles spawn with max vehcle settings
 Config.CustomAIPlates = 'ESX.A111' -- Custom plates for AI vehicles 
--- a/[core]/es_extended/config.lua
+++ b/[core]/es_extended/config.lua
@@ -63,6 +63,8 @@ Config.RemoveHudCommonents = {
 	[22] = false, --HUD_WEAPONS
 }
 
+Config.EnableGetSharedObjectEvent = false -- This is not recommended, Enables old getSharedObject event.
+
 Config.SpawnVehMaxUpgrades = true -- admin vehicles spawn with max vehcle settings
 Config.CustomAIPlates = 'ESX.A111' -- Custom plates for AI vehicles 
 -- Pattern string format
--- a/[core]/es_extended/server/common.lua
+++ b/[core]/es_extended/server/common.lua
@@ -11,12 +11,15 @@ Core.RegisteredCommands = {}
 Core.Pickups = {}
 Core.PickupId = 0
 Core.PlayerFunctionOverrides = {}
-
 Core.playersByIdentifier = {}
 
-AddEventHandler("esx:getSharedObject", function()
-	local Invoke = GetInvokingResource()
-	print(("[^1ERROR^7] Resource ^5%s^7 Used the ^5getSharedObject^7 Event, this event ^1no longer exists!^7 Visit https://documentation.esx-framework.org/tutorials/tutorials-esx/sharedevent for how to fix!"):format(Invoke))
+AddEventHandler("esx:getSharedObject", function(cb)
+	if Config.EnableGetSharedObjectEvent then
+		cb(ESX)
+	else
+		local Invoke = GetInvokingResource()
+		print(("[^1ERROR^7] Resource ^5%s^7 Used the ^5getSharedObject^7 Event, this event ^1no longer exists!^7 Visit https://documentation.esx-framework.org/tutorials/tutorials-esx/sharedevent for how to fix! Alternativly you can enable the config option for it!"):format(Invoke))
+	end
 end)
 
 exports('getSharedObject', function()
--- a/[core]/es_extended/server/common.lua
+++ b/[core]/es_extended/server/common.lua
@@ -18,7 +18,7 @@ AddEventHandler("esx:getSharedObject", function(cb)
 		cb(ESX)
 	else
 		local Invoke = GetInvokingResource()
-		print(("[^1ERROR^7] Resource ^5%s^7 Used the ^5getSharedObject^7 Event, this event ^1no longer exists!^7 Visit https://documentation.esx-framework.org/tutorials/tutorials-esx/sharedevent for how to fix! Alternativly you can enable the config option for it!"):format(Invoke))
+		print(("[^1ERROR^7] Resource ^5%s^7 Used the ^5getSharedObject^7 Event, This Is NOT Recommended and is very bad for performance! Visit https://documentation.esx-framework.org/tutorials/tutorials-esx/sharedevent For More details"):format(Invoke))
 	end
 end)
 


[core]/es_extended/config.lua
```diff
@@ -63,6 +63,8 @@ Config.RemoveHudCommonents = {
 	[22] = false, --HUD_WEAPONS
 }
 
+Config.EnableGetSharedObjectEvent = false -- This is not recommended, Enables old getSharedObject event.
```
Comment (Mycroft-Studios) [inline [core]/es_extended/config.lua:66]:
```suggestion
Config.EnableGetSharedObjectEvent = false -- Warning: Enabling This WILL decrease your servers performance, use this at your own precaution.
```

[core]/es_extended/client/common.lua
```diff
@@ -6,7 +6,11 @@ if GetResourceState('ox_inventory') ~= 'missing' then
...
 
-AddEventHandler("esx:getSharedObject", function()
-	local Invoke = GetInvokingResource()
-	print(("[^1ERROR^7] Resource ^5%s^7 Used the ^5getSharedObject^7 Event, this event ^1no longer exists!^7 Visit https://documentation.esx-framework.org/tutorials/tutorials-esx/sharedevent for how to fix!"):format(Invoke))
+AddEventHandler("esx:getSharedObject", function(cb)
+	if Config.EnableGetSharedObjectEvent then
+		cb(ESX)
+	else
+		local Invoke = GetInvokingResource()
+		print(("[^1ERROR^7] Resource ^5%s^7 Used the ^5getSharedObject^7 Event, this event ^1no longer exists!^7 Visit https://documentation.esx-framework.org/tutorials/tutorials-esx/sharedevent for how to fix! Alternativly you can enable the config option for it!"):format(Invoke))
```
Comment (Mycroft-Studios) [inline [core]/es_extended/client/common.lua:14]:
```suggestion
		print(("[^1ERROR^7] Resource ^5%s^7 Used the ^5getSharedObject^7 Event, This Is NOT Recommended and is very bad for performance! Visit https://documentation.esx-framework.org/tutorials/tutorials-esx/sharedevent For More details"):format(Invoke))
```

[core]/es_extended/server/common.lua
```diff
@@ -11,12 +11,15 @@ Core.RegisteredCommands = {}
...
 
-AddEventHandler("esx:getSharedObject", function()
-	local Invoke = GetInvokingResource()
-	print(("[^1ERROR^7] Resource ^5%s^7 Used the ^5getSharedObject^7 Event, this event ^1no longer exists!^7 Visit https://documentation.esx-framework.org/tutorials/tutorials-esx/sharedevent for how to fix!"):format(Invoke))
+AddEventHandler("esx:getSharedObject", function(cb)
+	if Config.EnableGetSharedObjectEvent then
+		cb(ESX)
+	else
+		local Invoke = GetInvokingResource()
+		print(("[^1ERROR^7] Resource ^5%s^7 Used the ^5getSharedObject^7 Event, this event ^1no longer exists!^7 Visit https://documentation.esx-framework.org/tutorials/tutorials-esx/sharedevent for how to fix! Alternativly you can enable the config option for it!"):format(Invoke))
```
Comment (Mycroft-Studios) [inline [core]/es_extended/server/common.lua:21]:
```suggestion
		print(("[^1ERROR^7] Resource ^5%s^7 Used the ^5getSharedObject^7 Event, This Is NOT Recommended and is very bad for performance! Visit https://documentation.esx-framework.org/tutorials/tutorials-esx/sharedevent For More details"):format(Invoke))
```

Mycroft-Studios (changes requested):
a better indicator would be better :)
if its going to be done, it might as well be done properly :)

Kenshiin13:
> All servers using the old version are full of scripts that give the worst possible performance.

Can you elaborate on that? The servers I am working with run just fine, even the scripts that still use the old event. Sure, it's not optimal, I prefer the imports as well as they are much cleaner and easier to use, but I wouldn't suggest that they cause any performance issues.

Mycroft-Studios:
>  but I wouldn't suggest that they cause any performance issues.

but they do. the event is not only Async so therefore requires Waits and more threads but also creates race conditions. The import also syncs ESX.PlayerData to any resource using it. which means resources arent doing ESX.GetPlayerData() and causing Massive memory issues.

its not an issue that is noticeable on 1-2 scripts, but has a massive impact at scale and with FiveM now supporting 2048 slots, Scalability is VERY Important. it might work on your 120 resource server with 1-5 people just fine, but for a 300+ script server with 100s if not over 1000 players, it wont and will be an issue

Mycroft-Studios:
>  All of the servers I've been working with intentionally do not upgrade to 1.9 because of the missing event.

in my opinion that is great. Less copy paste servers. know u need to actually know the basics of coding before using. if you cannot work out "remove these lines of code and then add this line in", then you shouldnt be trying to make a server.

Kenshiin13:
> in my opinion that is great. Less copy paste servers. now u need to actually know the basics of coding before using. 

Hilarious toxicity, as expected from the dimwit ESX team. The issue is with escrow encrypted scripts, as stated by the OP.

Mycroft-Studios:
> There is a difference between theoratical performance issues and actual impact.

okay? and? its still bad code at the end of the day.

Flambaa:
Regardless, the export still sends requests to the server, sends back down data to the client. Client > Client and server > server doesn't matter anyway.

I work on one of the biggest servers in Australia which averages 250 players a night. 

Our older resources still use the the event, and we have not had a single problem with performance or any bugs.

Anything new we do or anytime we are editing a script we obviously go and change it to the export however it isn't something we sweat about.

@Mycroft-Studios It also isn't our responsibility to ensure people are doing the right thing, it clearly states the risks in the config. People will know it isn't recommended. 

If the performance difference your taking about is the 0.001 difference you always sweat over then i don't support your argument. As @Kenshiin13 said, there's a difference between Theoretical and practical performance differences.

Flambaa:
image

How so? it does the same thing. 

I don't think you have ever worked in a proper server environment where there are hundreds of people playing. Not 1 single performance issue has appeared for us, no hitches, no freezes etc. 

Prior to the event being removed I have never heard of performance problems either.

Mycroft-Studios:
> I don't think you have ever worked in a proper server environment where there are hundreds of people playing

and thats where u underestimate my experience in fivem :)

> Not 1 single performance issue has appeared for us, no hitches, no freezes etc.

just because its for u, doesnt mean its for everyone :)

Servers can vary massively

Flambaa:
I'm not underestimating. 

Don't expect people to update their ESX if you cant provide backwards compatibility.

CsokiHUN:
> I'm not underestimating.
> 
> Don't expect people to update their ESX if you cant provide backwards compatibility.

It is in everyone's own interest to upgrade their server for better performance. We can't do continuous compatibility because we sometimes have to make big changes to old systems. There is also a tutorial on how to upgrade sharedObject so if you can't do it, don't run a server or use ESX.

Kenshiin13:
> There is also a tutorial on how to upgrade sharedObject so if you can't do it, don't run a server or use ESX.

Are you trying to be stupid or is that just your natural instinct?

Flambaa:
> It is in everyone's own interest to upgrade their server for better performance. We can't do continuous compatibility because we sometimes have to make big changes to old systems. There is also a tutorial on how to upgrade sharedObject so if you can't do it, don't run a server or use ESX.

You realise there is barely, if any performance difference with the export and event (server > server, client > client (Exports and events will do the same thing)).  

I seriously don't understand why are you sweating over this so much. Its just an optional config. 

I'm also very interesting in seeing the so called performance differences, with actual proof. Not word of mouth. :)

Flambaa:
Went and did it myself 

image

image

mcNuggets1:
If we just stopped paying hundreds of dollars for encrypted resources, that suck ass, this would not be an issue.
Escrow fucked this whole community over and we all know it, yet blame it onto the ESX devs, because they want to move forward with their FREE and OPEN SOURCE product.

If you guys are so interested in keeping ESX up to date, just add the export yourself + remove the warnings.
It's not much work. It's simple.

However if we're that adamant about giving an optrion to remove warnings, we might aswell completely remove the warning, as it then serves no purpose anymore.

CsokiHUN:
> Went and did it myself

The results you measured are not completely realistic, because for the event we have to wait (e.g. promise) until the ESX object is ready for use, but for the export we get the object immediately. If you take this into account the difference between the two is about **4-6ms**, yes it's not much but you have to take it into account.

Mycroft-Studios:
> Went and did it myself

your numbers are completely wrong since you fail to take into account how you actually use the event.

image
image

Using this code (which is the correct usage of the event), it shows that took on average (on my i7-7th Gen CPU, 64 ram Dedicated server) 17.96ms for the ESX object to be gotten.

However, it would be unfair for me not to use the export time so:
image
image

exact same testing method, but using the correct way to use the import, and i got a whooping 0.09ms after Rounding to 2 decimal places. that is an Extra 17.06ms added on to the server for literally no reason.

The import is a harder 1 to measure, so i messaged it manually using this code:
image

And i then restarted the server over and over to collect the data.
And i got a consistent 0.07/0.08ms.

So Yes, it is FAR slower

Mycroft-Studios:
If i use the 1000 number that you used for the loop on the event and export.

i get these numbers:
**event**: 51.22 ms
**export**: 0.29 ms

That is an extra 50.93ms on average

Mycroft-Studios:
Dont forget , this is on a high end dedicated Server. Most users Will **not** have that luxury so it will also be even slower on a standard Users server.

Mycroft-Studios:
So to finish it off, no, its not a 0.01 or even a 0.1 ms difference, it will legit affect boot times and Client Startups. and thats not even mentioning the Memory issues from using ESX.GetPlayerData() every time, which fucks up client memory usage massively

Mycroft-Studios:
[mycroft_benchmarks.zip](https://github.com/esx-framework/esx_core/files/11004940/mycroft_benchmarks.zip)

for those interested, you can test it yourself if you would like to validate my results

TheFantomas:
Hi, so I'd like to make a comment.

Thank you Mycroft for your willingness to measure the values. As we can see the potential performance of the event is not quite good rather bad 😄 I think 50ms extra is already a good reason not to bring this event back. 

If someone really needs it, can they put the event back into esx or not? Although I don't recommend it. With normal scripts it is not that hard to change an event to import or export. 

For escrowed stuff it's not so good but either you can agree with the creator or you are using some very outdated script that is no longer supported and it's better to try to find a developer who will rewrite/remake it 😄 

From my point of view, I don't want to support such an outdated non optimized thing and I probably won't. We will wait for the comments of the others and we will see. I will not decide this alone.

Best regards
Legacy Management
@TheFantomas

Benzo00:
I have reviewed the comments and decided to close this pr, due to that the event was removed for a reason, and as u see in Mycrofts comment there is a big performance difference and very noticeable. So using this old event is not something that should be done, update your scripts and do it the right way, the best way, the most performant way, instead of using old stuff, keep in mind when we remove something it’s not because we like to remove things, it’s because we have a valid reason todo so, like you’ve seen with this event

Benzo00:
Thanks everyone for their engagement and comments and benchmarks!

Gellipapa:
@Flambaa @Kenshiin13 
Hi! There are also huge changes to come, which is exactly what you want - a system that is up to date, fast and not least easy to use. 

About the config in short: The reason why we won't provide the config is precisely because everyone would just see a bug and change it, so what we do is worthless. We made the decision a while ago to implement it as a warning and the next step will be to keep the export-only solution.  

In the future there really will be a lot of breaking changes but we have to make this decision because we can't optimize by always referring to the old code. 

I say stick with us and together we will make the most stable and best ESX :)

Flambaa:
> @Flambaa @Kenshiin13 
> 
> Hi! There are also huge changes to come, which is exactly what you want - a system that is up to date, fast and not least easy to use. 
> 
> 
> 
> About the config in short: The reason why we won't provide the config is precisely because everyone would just see a bug and change it, so what we do is worthless. We made the decision a while ago to implement it as a warning and the next step will be to keep the export-only solution.  
> 
> 
> 
> In the future there really will be a lot of breaking changes but we have to make this decision because we can't optimize by always referring to the old code. 
> 
> 
> 
> I say stick with us and together we will make the most stable and best ESX :) 



If your going to repeat what someone else just said don't even bother writing this up

Gellipapa:
I just described the future for you, either you get used to it or you run away :)

> > @Flambaa @Kenshiin13
> > Hi! There are also huge changes to come, which is exactly what you want - a system that is up to date, fast and not least easy to use.
> > About the config in short: The reason why we won't provide the config is precisely because everyone would just see a bug and change it, so what we do is worthless. We made the decision a while ago to implement it as a warning and the next step will be to keep the export-only solution.
> > In the future there really will be a lot of breaking changes but we have to make this decision because we can't optimize by always referring to the old code.
> > I say stick with us and together we will make the most stable and best ESX :)
> 
> If your going to repeat what someone else just said don't even bother writing this up

Flambaa:
> I just described the future for you, either you get used to it or you run away 🚀

Pretty sure the conversation ended after my comment. Stop waffling

Scullyy:
```lua
ESX = exports.es_extended:getSharedObject()

RegisterNetEvent('esx:getSharedObject', function(cb)
    cb(ESX)
end)
```

einfachyannik (commented):
It just helps perfectly

PR #938, opened 2023-03-11
chore(esx_multicharacter/locales/nl.lua): Update Dutch Locales

Arctos2win:
Fixing nl.lua from esx_multichar


TheFantomas:
Copy of #940

PR #946, opened 2023-03-14
Create cn.lua

442516:
Simplified Chinese translation


TheFantomas:
Hi, we have our own translator for this language.

Even so, you are missing a few lines in your translation and this is one of the other reasons why this pull request cannot be accepted.

Thanks anyway for trying to help with the development of the ESX Framework.

With respect and regards
Legacy Management
@TheFantomas

PR #947, opened 2023-03-14
feat(esx/server/classes/player): Add Job Duty

Thekuca:


CsokiHUN (approved):
Looks good but, do we add it to the database save?

TheFantomas:
> Looks good but, do we add it to the database save?

Maybe we can add it to the database. Is good suggestion from you :D

Mycroft-Studios:
this a bad idea, its basically the exact same system that was in before and benzo had to remove. for this to work it has to be added into all resources otherwise there is no point in it, just adding a variable in doesn't add the functionality, that part lies in the jobs themselves

TheFantomas:
> this a bad idea, its basically the exact same system that was in before and benzo had to remove. for this to work it has to be added into all resources otherwise there is no point in it, just adding a variable in doesn't add the functionality, that part lies in the jobs themselves

This will be integrated in the next release so 1.9.5. It will be integrated into all jobs, don't worry :D

Mycroft-Studios:
> > this a bad idea, its basically the exact same system that was in before and benzo had to remove. for this to work it has to be added into all resources otherwise there is no point in it, just adding a variable in doesn't add the functionality, that part lies in the jobs themselves
> 
> This will be integrated in the next release so 1.9.5. It will be integrated into all jobs, don't worry :D

if its integrated in thats fine, but the PR alone is a bad idea :)

PR #952, opened 2023-03-17
state Bag Support For Weapons removing the loop huge optimization

Fusionfivem:


Mycroft-Studios:
umm, where are u setting this state bag?

Fusionfivem:
uhh the client side may have not saved one moment

PR #955, opened 2023-03-18
fix(es_extended/server/functions.lua): Added model argument to callback

TheFantomas:


[core]/es_extended/server/functions.lua
```diff
@@ -262,7 +262,7 @@ function ESX.GetVehicleType(model, player, cb)
   ESX.TriggerClientCallback(player, "esx:GetVehicleType", function(vehicleType)
     Core.vehicleTypesByModel[model] = vehicleType
```
Comment (Gellipapa) [inline [core]/es_extended/server/functions.lua:263]: How to get model value? If you add  argument to callback, you need init function part. 

`
ESX.TriggerClientCallback(player, "esx:GetVehicleType", function(vehicleType, model)
    Core.vehicleTypesByModel[model] = vehicleType or Core.vehicleTypesByModel[vehicleType.model] = vehicleType
    cb(vehicleType)
  end, model)
`
IDK how the customer callback works I have never seen this just a hint.

CsokiHUN (approved):
I left that out, thanks for your fix

PR #957, opened 2023-03-18
feat(esx_multicharacter): Use dummy identifier for multicharacter when in fxdk mode

donxp:
Use dummy identifier in the multicharacter resource when determining player identifier. This syncs the functionality to what already is happening in `es_extended/server/functions.lua` `ESX.GetIdentifier` method.

Without this, multicharacter resource does not detect existing characters when running under fxdk and will always prompt the user to create a new one, which will fail on the INSERT query due to duplicate primary keys.


CsokiHUN (approved):
Thanks.

PR #958, opened 2023-03-19
Update server.cfg

Arctos2win:
stopping resources that will break the core

### DIFF
--- a/server.cfg
+++ b/server.cfg
@@ -56,3 +56,12 @@ ensure [core]
 
 ## ESX Addons
 stop basic-gamemode #remove this if you don't want to use multicharacter
+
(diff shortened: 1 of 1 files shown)


TheFantomas:
Hello, from my point of view, this gadget is a bit unnecessary.

Directly in es_extended, when you use such a resource, it will print that you're using the wrong resource..
You can see it here: https://github.com/esx-framework/esx_core/blob/main/%5Bcore%5D/es_extended/client/main.lua#L633-L648

Therefore this pr will be closed.
Anyway, thank you for contributing to the improvement of ESX.

Yours sincerely
Legacy Management
**Fantík**

PR #861, opened 2023-01-25
feat(server/functions): modify ESX.Jobs table and its related database tables on runtime

mahanmoulaei:
#### Add server-side functions to modify the core jobs table and database without the need to restart the server
Example Usage:

**ESX.AddJob**
```lua
-- adding single job
local job = {
    name = "JOB_NAME",
    label = "JOB_LABEL",
    whitelisted = true or false,
    grades = {
        ["0"] = {name = "trainee", label = "Trainee", salary = 100, skin_male = {}, skin_female = {}},
        ["1"] = {name = "experienced", label = "Experienced", salary = 150, skin_male = {}, skin_female = {}},
        ["2"] = {name = "boss", label = "Manager", salary = 200, skin_male = {}, skin_female = {}},
    }
}
ESX.AddJob(job)

-- adding multiple jobs
local jobs = {
    {
        name = "JOB_1_NAME",
        label = "JOB_1_LABEL",
        whitelisted = true or false,
        grades = {
            ["0"] = {name = "trainee", label = "Trainee", salary = 100, skin_male = {}, skin_female = {}},
            ["1"] = {name = "experienced", label = "Experienced", salary = 150, skin_male = {}, skin_female = {}},
            ["2"] = {name = "boss", label = "Manager", salary = 200, skin_male = {}, skin_female = {}},
        }
    },
    {
        name = "JOB_2_NAME",
        label = "JOB_2_LABEL",
        whitelisted = true or false,
        grades = {
            ["0"] = {name = "trainee", label = "Trainee", salary = 100, skin_male = {}, skin_female = {}},
            ["1"] = {name = "experienced", label = "Experienced", salary = 150, skin_male = {}, skin_female = {}},
            ["2"] = {name = "boss", label = "Manager", salary = 200, skin_male = {}, skin_female = {}},
        }
    }
}
ESX.AddJob(jobs)
```
<hr>

**ESX.UpdateJob**
```lua
-- updating single job
-- *for optional attributes => if provided it will update it, and if not provided it will use the previously saved values*
local job = {
    name = "JOB_NAME", -- mandatory
    label = "JOB_LABEL", -- optional
    whitelisted = true or false, -- optional
    grades = { -- optional
        ["0"] = {name? = "trainee", label? = "Trainee", salary? = 100, skin_male? = {}, skin_female? = {}}, -- all of the attributes are optional
[...]

PR #960, opened 2023-03-20
feat(es_extended/server/functions.lua): ESX.CreateJob(tb)

wasabirobby:
Create a job during runtime. If necessary will write to SQL the data if not already present

### DIFF
--- a/[core]/es_extended/server/functions.lua
+++ b/[core]/es_extended/server/functions.lua
@@ -313,38 +313,35 @@ function ESX.DiscordLogFields(name, title, color, fields)
   })
 end
 
-function ESX.CreateJob(tb)
-  -- tb (table) containing: name (string), label (string) and grades (table)
-  -- grades (table) contains: grade (int), name (string), label (string), salary (int)
-  
-  if tb == nil or next(tb) == nil then
-      return print('[^3WARNING^7] paramter must be a table.')
+--- Create Job at Runtime
+--- @param name string
+--- @param label string
+--- @param grades table
+function ESX.CreateJob(name, label, grades)
+  if not name then
+      return print('[^3WARNING^7] missing argument `name(string)` while creating a job')
   end
   
-  if tb.name == nil then
-      return print('[^3WARNING^7] table must contain name(string) for the job name')
+  if not label then
+      return print('[^3WARNING^7] missing argument `label(string)` while creating a job')
   end
   
-  if tb.label == nil then
-      return print('[^3WARNING^7] table must contain label(string) for the job name label')
-  end
-  
-  if tb.grades == nil or next(tb.grades) == nil then
-      return print('[^3WARNING^7] table must contain grades(table)!')
+  if not grades or not next(grades) then
+      return print('[^3WARNING^7] missing argument `grades(table)` while creating a job!')
   end
 
   local parameters = {}
-  local job = {name = tb.name, label = tb.label, grades = {}}
+  local job = {name = name, label = label, grades = {}}
 
-  for k,v in pairs(tb.grades) do
-      job.grades[tostring(v.grade)] = {job_name = tb.name, grade = v.grade, name = v.name, label = v.label, salary = v.salary, skin_male = {}, skin_female = {}}
-      parameters[#parameters + 1] = { tb.name, v.grade, v.name, v.label, v.salary}
+  for k,v in pairs(grades) do
+      job.grades[tostring(v.grade)] = {job_name = name, grade = v.grade, name = v.name, label = v.label, salary = v.salary, skin_male = {}, skin_female = {}}
+      parameters[#parameters + 1] = { name, v.grade, v.name, v.label, v.salary}
   end
 
-  MySQL.insert('INSERT IGNORE INTO jobs (name, label) VALUES (?, ?)', {tb.name, tb.label})
+  MySQL.insert('INSERT IGNORE INTO jobs (name, label) VALUES (?, ?)', {name, label})
   MySQL.prepare('INSERT INTO job_grades (job_name, grade, name, label, salary) VALUES (?, ?, ?, ?, ?)', parameters)
   
-  ESX.Jobs[tb.name] = job
+  ESX.Jobs[name] = job
 end
 
 function ESX.RefreshJobs()


[core]/es_extended/server/functions.lua
```diff
@@ -313,6 +313,25 @@ function ESX.DiscordLogFields(name, title, color, fields)
   })
 end
 
+function ESX.CreateJob(tb)
```
Comment (Gellipapa) [inline [core]/es_extended/server/functions.lua:316]: tb handle nil

[core]/es_extended/server/functions.lua
```diff
@@ -313,6 +313,25 @@ function ESX.DiscordLogFields(name, title, color, fields)
...
+  -- tb (array) containing: name (string), label (string) and grades (table)
+  -- grades (table) contains: grade (int), name (string), label (string), salary (int)
+  local job = {}
+  MySQL.Async.execute('INSERT IGNORE INTO jobs (name, label) VALUES (@name, @label)', {['name'] = tb.name, ['label'] = tb.label})
+  job = {name = tb.name, label = tb.label, grades = {}}
+  for k,v in pairs(tb.grades) do
+    job.grades[tostring(v.grade)] = {job_name = tb.name, grade = v.grade, name = v.name, label = v.label, v.salary, skin_male = {}, skin_female = {}}
+    MySQL.Async.execute('INSERT IGNORE INTO job_grades (job_name, grade, name, label, salary) VALUES (@job_name, @grade, @name, @label, @salary)', {
```
Comment (Gellipapa) [inline [core]/es_extended/server/functions.lua:324]: I think ox can multiple insert if you add full table not needed in for loop.

Gellipapa (commented):
Please check this.

mahanmoulaei:
While I agree that the core should have job-modifier functions, I have a question for ESX team:

How come a [comparable pull request](https://github.com/esx-framework/esx_core/pull/861), which included more extensive functions and type-checking, was closed under the justification that "after professional consultation we would rather rewrite the entire job system so that it is in the config and is completely removed from the database," when a similar pull request with less comprehensive functions for modifying jobs is now being reviewed after just two months?

Gellipapa:
> 

This is a really good question. 
Your solution was so detailed I'm sure no one would use it in runtime. The way I looked at it, it's a simpler solution, but it doesn't really make any sense. 

Why should the job be set at runtime? 

Thanks for letting me know, I'm looking at the pr's quite late at some point and would rather look at the code than the whole function.

Gellipapa:
@wasabirobby Can you tell me some use cases for which this would be useful? Can you tell me some use cases for which this would be useful? To add runtime jobs to the DB.

mahanmoulaei:
> Your solution was so detailed

Don't include a function because it's too detailed hahaha

> I'm sure no one would use it in runtime.

It will be used. Just because you & other ESX team members don't have any use case for a function, doesn't mean others won't use it too! Don't limit others to only see inside the box as you do. Other cores such as QB has it(however it's broken), me, wasabi, [jaksam1074](https://forum.cfx.re/t/jobs-creator-5-0-esx-qbcore-in-game-job-creation-menu-without-server-restart/2667762), [renzu](https://github.com/renzuzu/renzu_jobs), [kmack710](https://forum.cfx.re/t/free-qbcore-in-game-simple-job-creator/4815680), [DV-Scripts](https://forum.cfx.re/t/qb-paid-dv-jobs-job-creator/4857195), [kkyabe](https://forum.cfx.re/t/paid-qbcore-business-job-creator/4835888), [robbietrains](https://forum.cfx.re/t/free-qbcore-boppe-creator-dynamically-create-jobs-and-gangs-in-game/5004970), [paralyst_punch](https://forum.cfx.re/t/esx-free-base-script/4928858), [chief-tron ](https://forum.cfx.re/t/paid-qb-ct-jobcreator-easy-in-game-job-creation-wardrobe-on-off-duty-bossmenu-stash-location-setters/5018391), [guille](https://forum.cfx.re/t/release-free-guille-jobcreator/2897543), [ali_exacute](https://forum.cfx.re/t/esx-advanced-job-system-v1-1-0/2616104), more developers, and all those who download their resource intend to use such functionalities.

> Why should the job be set at runtime?

For the same reason `ESX.RefreshJobs()` was introduced. (like why would you want to refresh the job table if it's not going to be modified on runtime?)

> I'm looking at the pr's quite late at some point and would rather look at the code than the whole function.

I didn't get this sentence buddy

<hr>

I have already mentioned the use cases that **I could think** of within my PR, but I'm writing them here one more time:
- Adding a job to the core directly within external and third-party resources without the need to provide a SQL for it.

- Modifying all attributes of a job like changing its label, whitelisted, grade label, grade name, grade salary, grade skins for employee, adding more grades & etc on runtime without the need to go to the database and change the desired data and have a permitted player to execute /refreshjobs command or do a server restart.

- Make esx_society's boss menu more configurable by giving a job's boss the ability to modify ranks/grades label, salary, job skins and ...

- Update players' jobs on any change applied to ESX.Jobs table without the need to re-assign them the job(setJob) or have them to leave the server and join back again.

- There are a lot of resources out there that aim to create jobs and related grades which require a server restart or executing /refreshjob command for those modifications to take place

Gellipapa:
I understand what you're writing but I don't see the point of dealing with these at the core level because it doesn't make sense. So if you create a script and change something in the database, say add new jobs, why should you write it into CORE and call things via ESX function? Much more risk for a core and unnecessary. 

If I make a script I'll handle it according to my logic, it's not my goal to call the ESX.CreateJob() function and make it difficult to write standalone.

Because if there was one for ESX would this function really be used in most cases do you think? 

I think not precisely because everyone writes their own logic and since almost the database is built the same you can easily standalone shoulder the script and not have to look at whether it is in ESX for QB not or vice versa.

But convince me if you think it is.

Gellipapa:
So you have written a logic so complex that it is very risky to make it work well. Lots of SQL queries, lots of configuration options and relatively little error handling logic. 

If the code you write breaks at any point and everyone writes everything to it, that would be a pretty big problem. 

You'd have to first separate the code, then extract the code iterations, optimize it and then test it A-Z and submit it because you can't expect the complete logic you wrote to be fully error-proofed so it's a risk. 

We have to think carefully about what we accept because it may be useful on one side but if the risk is high it is not worth it. 

Anyway, the JOB handling logic will probably be rewritten because the current one is quite cumbersome, doesn't handle multijobs etc.

mahanmoulaei:
> why should you write it into CORE and call things via ESX function? Much more risk for a core and unnecessary.
If I make a script I'll handle it according to my logic, it's not my goal to call the ESX.CreateJob() function and make it difficult to write standalone.

Job attribute is a framework-related thing, not something that you could achieve as standalone. Also, why would you want to write standalone if you are using a framework? Being standalone and having a framework are kind of two opposite sides, so I don't understand why are you using ESX and want to write standalone?

> Because if there was one for ESX would this function really be used in most cases do you think?
I think not precisely because everyone writes their own logic and since almost the database is built the same you can easily standalone shoulder the script and not have to look at whether it is in ESX for QB not or vice versa.

Why would you want to handle players and their data through the core when you can just implement player handling system within every single script manually and independently?
The main goal of a framework is to centralize functions and handlers.
As another example, when you could write your own vehicle spawning function in every single resource manually, why would you include that function inside the framework?
- firstly, it makes it easy to use by calling just 1 line of the function name and its required parameter.
- secondly, it makes the vehicle function shared and consistence across all of your resources.
- thirdly, if you had to change the behavior and body of the vehicle spawning function, if you have it inside the core you could just easily modify 1 resource only; while if you have that function written manually in every single script, then it makes your life hard.

> So you have written a logic so complex that it is very risky to make it work well. Lots of SQL queries, lots of configuration options and relatively little error handling logic.
If the code you write breaks at any point and everyone writes everything to it, that would be a pretty big problem.

Do you wonder why my PR code logic looks so complex? Because literally, every single attribute and itheir types are being checked and if anything goes wrong, the execution of the function will be terminated returning a message describing the error that happened.

> You'd have to first separate the code, then extract the code iterations, optimize it and then test it A-Z and submit it because you can't expect the complete logic you wrote to be fully error-proofed so it's a risk.

How do you know I haven't done this in prior to marking my PR as ready to review? Did you even test it before assuming me not testing my code? How could I write a complex logic without testing it in the first place?

> We have to think carefully about what we accept

This is really a good thing to consider and I encourage ESX to do it more, however, this has not been the case a lot of times. Many accepted commit changes that are intended to improve existing codes actually end up making them worse. How many times have you all had to modify the VehicleProperties state bag change handler body when the code was not functioning correctly due to insufficient consideration prior to accepting code changes?
[This example](https://github.com/esx-framework/esx_core/blob/main/%5Bcore%5D/es_extended/client/main.lua#L260-L265) shows an unoptimized implementation of state bag change handlers by Fatnik(if I'm not mistaken) that appears to have been accepted without proper consideration, contradicting the claim that careful consideration is always taken before accepting commits.

<hr>


Lastly, I need to mention I said all of these not to request the team to accept my PR, but to try making things clear.
Moreover, please don't take anything I say as trash-talking or hate toward the ESX members and their capabilities.
Make ESX shine and great again

Gellipapa:
Thank you very much for writing this down. 

I first thought of the database part when I said that writing your own logic is a better choice. 

Of course, spawning a car is a different story because that's where I handle the additional parameters that make it all coherent.

I didn't claim that you didn't test it, I just described how a PR should be cast if you tested it then you're doing it right. Personally I didn't test your code I just reviewed it. That's my mistake in fact.

Yes it does happen that we make mistakes and accept code that doesn't always work well, but we try to reduce the number of mistakes. 

Yes, we can expect very big changes in the future from an ESX point of view, but more on that later... 

I understand what your goal is, but I also don't want the not so savvy to call createJob in Thread or not use it the way it should be used. 

We also have to take into account how easy it is to use the system. We have no secret goal to create a system that satisfies the advanced user.Our goal is to create a system that meets the needs of both advanced and novice developers. 
You will learn more about this soon. 

We're here and we're working hard, and thank you for alerting us to bugs and helping us improve ESX. 
No offense intended, I'd probably be annoyed if I did something for days and it didn't get done. 
I understand your point of view too but as I said we need to look at PRs from more angles. 
These will be looked at more in the future to make the code more stable.

mahanmoulaei:
I'm sure all ESX members are trying their best to make the resources better every day, but as a suggestion, maybe a different approach and point of view, could bring even more achievement for the team.

I look forward to hearing more about the upcoming changes in ESX and its future.

Thank you for your answers.

wasabirobby:
Did not intend for it to become an argument. Anyway, I have been moving homes, apologies for the delay. I resolved the previous issues.

CsokiHUN (approved):

PR #978, opened 2023-03-24
refactor: slight optimization

MoskalykA:
optimization is explained by this article:  https://www.builder.io/blog/double-equal-vs-triple-equal

### DIFF
--- a/[core]/esx_context/index.html
+++ b/[core]/esx_context/index.html
@@ -269,7 +269,7 @@
           }
 
           if (ele.input) {
-            if (ele.inputType == "radio") {
+            if (ele.inputType === "radio") {
               for (let j = 0; j < ele.inputValues.length; j++) {
                 let v = ele.inputValues[j];
                 let container = CreateElement("label", "container", div);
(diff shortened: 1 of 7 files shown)


CsokiHUN (approved):
Looks good, thanks

CsokiHUN (approved):

[core]/esx_context/index.html
```diff
@@ -390,8 +390,8 @@
       });
 
       window.addEventListener("keydown", (e) => {
-        if (e.key == "Escape" || e.key == "Backspace") {
-          if (e.target.tagName.toLowerCase() == "input") {
+        if (e.key === "Escape" || e.key === "Backspace") {
```
Comment (jrence) [inline [core]/esx_context/index.html:393]: may i know why ===  or miss type  ? for are they same???

Comment (MoskalykA) [reply [core]/esx_context/index.html:393]: > may i know why === or miss type ? for are they same???

I didn't understand your comment

Comment (Thekuca) [reply [core]/esx_context/index.html:393]: === is faster since == checks for the type of the variable and then compares while === checks if they are equal

jrence (approved):

PR #987, opened 2023-03-29
Delete [core]/async directory

Flambaa:


CsokiHUN:
why?

Flambaa:
> why?

it's not used anymore

thelindat:
> > why?
> 
> it's not used anymore

and if other resources use it?

Flambaa:
> > > why?
> > 
> > 
> > it's not used anymore
> 
> and if other resources use it?

I should say, not used by any ESX scripts. If other resources need it, there are copies of it floating around github. Could even upload async as a separate repo if that helps.

jrence:
i think we should keep it for a while..

Flambaa:
> i think we should keep it for a while.. 

Why?

TheFantomas:
Hi, 

I think we can make a compromise. I don't think this script needs to be directly in the core folder, I'd rather upload it to github as a separate script and archive it.

There is no reason for it to be in the core folder if it is not used by any of our scripts.

Thank you all for participating in this conversation.

Yours sincerely and respectfully
Legacy Management
@TheFantomas

PR #1004, opened 2023-04-02
feat(es_extended/server/functions.lua): "merge" args type in ESX.RegisterCommand

GurraG004:
added so you can merge the args for reports command and other functions when you are sending a message with spaces. Needs to be the last args and validate set to false. / Fores


Thekuca (approved):
Nice Feat

CsokiHUN (approved):

PR #1009, opened 2023-04-05
fix(esx/ox_inventory): trigger esx:setAccountMoney when syncing inventory

thelindat:
Change was excluded from https://github.com/esx-framework/esx_core/commit/2851032034f93eb48863b909f2bfdb9c7b3fa76f


Benzo00:
Mycroft

CsokiHUN (approved):
Thanks

PR #1020, opened 2023-04-16
Added Command to test all notifyes at once

Sooeeren:


TheFantomas:
I think thats not needed.

PR #1022, opened 2023-04-19
Add Support for Drift Tyres

Kuuzoo:
Integrate them into the VehicleProperties Functions, and make sure that they return true on the check so there should be no problems with older builds.

Thx to mycroft for the idea with the type <3

### DIFF
--- a/[core]/es_extended/client/functions.lua
+++ b/[core]/es_extended/client/functions.lua
@@ -627,6 +627,11 @@ function ESX.Game.GetVehicleProperties(vehicle)
         end
     end
 
+    local driftTyresEnabled = false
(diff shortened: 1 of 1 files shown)


[core]/es_extended/client/functions.lua
```diff
@@ -627,6 +627,11 @@ function ESX.Game.GetVehicleProperties(vehicle)
         end
     end
 
+    local driftTyresEnabled = false
+    if type(GetDriftTyresEnabled(vehicle) == boolean) and GetDriftTyresEnabled(vehicle) then
```
Comment (CsokiHUN) [inline [core]/es_extended/client/functions.lua:631]:
```suggestion
    if type(GetDriftTyresEnabled(vehicle) == "boolean") and GetDriftTyresEnabled(vehicle) then
```

CsokiHUN (approved):

Thekuca (approved):

PR #1031, opened 2023-04-28
feat(es_extended): add Await for callbacks

CsokiHUN:


Rav3n95 (approved):

Gellipapa (approved):
So clean i like that. Thank you for new await feature.

Thekuca (approved):

PR #1035, opened 2023-05-01
feat(xPlayer): add hasMeta class function

CENSOR1337:
Add a function to check existence of metadata table key.

### DIFF
--- a/[core]/es_extended/server/classes/player.lua
+++ b/[core]/es_extended/server/classes/player.lua
@@ -681,6 +681,18 @@ function CreateExtendedPlayer(playerId, identifier, group, accounts, inventory, 
 		Player(self.source).state:set('metadata', self.metadata, true)
 	end
 
+    function self.hasMeta(index)
+        if not index then
+            return print(("[^1ERROR^7] xPlayer.hasMeta ^5%s^7 is Missing!"):format(index))
+        end
+
+        if type(index) ~= 'string' then
+            return print("[^1ERROR^7] xPlayer.hasMeta ^5index^7 should be ^5string^7!")
+        end
+
+        return self.metadata[index] ~= nil
+    end
+
 	for fnName,fn in pairs(targetOverrides) do
 		self[fnName] = fn(self)
 	end
--- a/[core]/es_extended/server/classes/player.lua
+++ b/[core]/es_extended/server/classes/player.lua
@@ -615,7 +615,7 @@ function CreateExtendedPlayer(playerId, identifier, group, accounts, inventory, 
 
 				return self.metadata[index]
 			else
-				return print(("[^1ERROR^7] xPlayer.getMeta ^5%s^7 not exist!"):format(index))
+				return
 			end
 
 		end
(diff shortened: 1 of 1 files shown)


TheFantomas:
**Hello,**

The previous ESX Legacy leader and I agreed that we didn't know where this could be used. So I'm asking you where it could be used? If you give a reason and we find it understandable your pull request will be accepted/merged.

Sincerely and respectfully
Management/Leader of Legacy Team
**Fantík**

CENSOR1337:
This can be used to verify the existence of a key before it is used or accessed, ex.
```lua
if not (xPlayer) then return false end
local expireTime = xPlayer.getMeta("csjob_bonus_multiplier_timeend")
if not (expireTime) then return false end
local bExpired = expireTime < os.time()
if (bExpired) then
    xPlayer.clearMeta("csjob_bonus_multiplier_timeend")
end
return not (bExpired)
```

This was used to store item bonus expiration time when it expires then delete the key entry,
as you can see if a player has no index will print an error message, 
```lua
print(("[^1ERROR^7] xPlayer.getMeta ^5%s^7 not exist!"):format(index))
```
because it is always a good idea to clear unused data.

so i think it's a good practice to check entry before it being access
```lua
if not (xPlayer.hasMeta("csjob_bonus_multiplier_timeend")) then return end
```

Thank you.

TheFantomas:
> This can be used to verify the existence of a key before it is used or accessed, ex.
> 
> ```lua
> if not (xPlayer) then return false end
> local expireTime = xPlayer.getMeta("csjob_bonus_multiplier_timeend")
> if not (expireTime) then return false end
> local bExpired = expireTime < os.time()
> if (bExpired) then
>     xPlayer.clearMeta("csjob_bonus_multiplier_timeend")
> end
> return not (bExpired)
> ```
> 
> This was used to store item bonus expiration time when it expires then delete the key entry, as you can see if a player has no index will print an error message,
> 
> ```lua
> print(("[^1ERROR^7] xPlayer.getMeta ^5%s^7 not exist!"):format(index))
> ```
> 
> because it is always a good idea to clear unused data.
> 
> so i think it's a good practice to check entry before it being access
> 
> ```lua
> if not (xPlayer.hasMeta("csjob_bonus_multiplier_timeend")) then return end
> ```
> 
> Thank you.

I think you can do if not xPlayer.getMeta('whatever') then return end and its the same or?

CENSOR1337:
Yes it provides the same result but with
```lua
xPlayer.getMeta('whatever')
``
will print an error message on the console 

As you can see I can use this line and it works as expected except the console will be filled with and error message
```lua
local expireTime = xPlayer.getMeta("csjob_bonus_multiplier_timeend")
if not (expireTime) then return false end
```

CENSOR1337:
```lua
function xPlayer:getMeta(index, subIndex)
    if index then
        ....
        else
            return print(("[^1ERROR^7] xPlayer.getMeta ^5%s^7 not exist!"):format(index))
        end
    end

    return self.metadata
end
```
As you can see when the entry does not exist will print an error message as I provided above

Thank you

mahanmoulaei:
instead of a new function, the return print error in getMeta should be swapped with `nil`

CENSOR1337:
> instead of a new function, the return print error in `getMeta('something')` should be swapped with `nil` if the index doesn't exist. simple as that

If that was a concern about adding a new function yes that will do.
I can edit my pr depends on protect teams

Thank you.

TheFantomas:
> > instead of a new function, the return print error in `getMeta('something')` should be swapped with `nil` if the index doesn't exist. simple as that
> 
> If that was a concern about adding a new function yes that will do. I can edit my pr depends on protect teams
> 
> Thank you.

Like @mahanmoulaei said :D

CENSOR1337:
I will open another PR to match title 
Thanks

TheFantomas:
Okey

PR #1037, opened 2023-05-03
refactor(xPlayer): delete getMeta error message

CENSOR1337:
getMeta no longer prints an error message on the index did not exist condition.

### DIFF
--- a/[core]/es_extended/server/classes/player.lua
+++ b/[core]/es_extended/server/classes/player.lua
@@ -615,7 +615,7 @@ function CreateExtendedPlayer(playerId, identifier, group, accounts, inventory, 
 
 				return self.metadata[index]
 			else
-				return
+				return Config.EnableDebug and print(("[^1ERROR^7] xPlayer.getMeta ^5%s^7 not exist!"):format(index)) or nil
 			end
 
 		end
--- a/[core]/es_extended/server/classes/player.lua
+++ b/[core]/es_extended/server/classes/player.lua
@@ -604,7 +604,11 @@ function CreateExtendedPlayer(playerId, identifier, group, accounts, inventory, 
 
                 for i = 1, #subIndex do
                     local key = subIndex[i]
-                    returnValues[key] = self.getMeta(index, key)
+                    if (type(key) == "string") then
+                        returnValues[key] = self.getMeta(index, key)
+                    else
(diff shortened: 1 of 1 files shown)


CsokiHUN:
why?

NikartikTheBear:
I believe he meant why would you remove that

Gellipapa:
@CENSOR1337 Hi! Where use hasMeta? 

So not being used anywhere but just an option no one will use it but rely on getMeta() ~= nil or similar checks so I don't think it makes sense to implement it. 

Instead, you could have something that if the server is in debug mode then print returns getMeta with nil of course but then you have debug print which is useful for development.

CENSOR1337:
> @CENSOR1337 Hi! Thank you, I think we can refactor this code to be fully recursive, so the subIndex part could be replaced by a recursive function call and the code wouldn't be so complex. I know you didn't do this, but you are currently modifying this part of the code.

I have made a change base on your suggestion and maintenance a little bit of code flow, to make it easier to read.
Let me know what you think, Thanks.

[core]/es_extended/server/classes/player.lua
```diff
@@ -579,49 +579,43 @@ function CreateExtendedPlayer(playerId, identifier, group, accounts, inventory,
...
-			if type(index) ~= "string" then
-				return print("[^1ERROR^7] xPlayer.getMeta ^5index^7 should be ^5string^7!")
-			end
+        if type(index) ~= "string" then
+            return print("[^1ERROR^7] xPlayer.getMeta ^5index^7 should be ^5string^7!")
+        end
 
-			if self.metadata[index] then
+        local metadata = self.metadata[index]
+        if (metadata == nil) then
```
Comment (Gellipapa) [inline [core]/es_extended/server/classes/player.lua:590]: if not metadata then better.

[core]/es_extended/server/classes/player.lua
```diff
@@ -579,49 +579,43 @@ function CreateExtendedPlayer(playerId, identifier, group, accounts, inventory,
...
-							end
-						end
+            if (_type == "table") then
+                local returnValues = {}
 
-						return returnValues
-					end
+                for i = 1, #subIndex do
+                    local key = subIndex[i]
+                    returnValues[key] = self.getMeta(index, key)
```
Comment (Gellipapa) [inline [core]/es_extended/server/classes/player.lua:607]: key is table break code. You should handle only string.

Gellipapa (changes requested):
Nice so clean better than before version.

Gellipapa (approved):
Good job, thank you for help.

PR #1042, opened 2023-05-10
fix: multicharacter commands + locales formatting

0xDEMXN:


CsokiHUN (approved):
Looks good. Please resolve conflicts.

0xDEMXN:
Ready to merge

0xDEMXN:
Resolved conflicts again, can you merge this if it's ok? It's a pain to bulk edit apostrophes every time a locale is changed.
@CsokiHUN

[core]/esx_multicharacter/server/commands.lua
```diff
@@ -1,28 +1,16 @@
 ESX.RegisterCommand('setslots', 'admin', function(xPlayer, args, showError)
-	local slots = MySQL.scalar('SELECT `slots` FROM `multicharacter_slots` WHERE identifier = ?', {
-		args.identifier
-	})
-
-	if not slots then
-		MySQL.update('INSERT INTO `multicharacter_slots` (`identifier`, `slots`) VALUES (?, ?)', {
+		MySQL.insert('INSERT INTO `multicharacter_slots` (`identifier`, `slots`) VALUES (?, ?) ON DUPLICATE KEY UPDATE `slots` = VALUES(`slots`)', {
```
Comment (Gellipapa) [inline [core]/esx_multicharacter/server/commands.lua:2]: Why remove whole code?

Gellipapa (changes requested):

[core]/esx_multicharacter/server/commands.lua
Comment (0xDEMXN) [reply [core]/esx_multicharacter/server/commands.lua:2]: Because duplicate entries are managed inside the mysql insert function, it's not needed anymore.

0xDEMXN:
> @0xDEMXN Hi! Thank you for formatting all locales but i think we can do it automatically so there is no risk that some error occurred. (typo etc)

It was meant for end users more than contributors who sometimes want to change a locale by himself and doesn't know how to escape a single quote intended as apostrophe.

Gellipapa:
> 
Of course I understand, and thank you very much, but there are so many new translations all the time that we need to automate this process.

[core]/esx_multicharacter/server/commands.lua
Comment (Gellipapa) [reply [core]/esx_multicharacter/server/commands.lua:2]: Yes it's okey but you can't seprate now add and update notify. You removed slotsedit notify. This is not correct because the player does not know if he has just added a slot or if he has already modified an existing slot.

BUT
If we don't look at this in the future we will have to delete the slotsedit key from all locales and yes this will be the right direction thanks again.

Gellipapa (changes requested):

[core]/esx_multicharacter/server/commands.lua
Comment (0xDEMXN) [reply [core]/esx_multicharacter/server/commands.lua:2]: The command is literally 'setslots': you set the slots you want with a given number, what's the point of knowing if you are adding or updating a slot?

Comment (Gellipapa) [reply [core]/esx_multicharacter/server/commands.lua:2]: @0xDEMXN Hi! Yes nothing, please delete slotsedit keys all locales and i can merge. Thank you.

Gellipapa:
@0xDEMXN Thank you for this fixes, i resolved conflict and merged.

PR #1043, opened 2023-05-12
Added option to externally cancel progress bar

FekyDEV:
Personally, I think this feature should have been here a long time ago. I hope that adding this code will help players in the future and if anyone here has a suggestion for a code change feel free to comment.


CsokiHUN (approved):
good idea, thanks

PR #1044, opened 2023-05-13
Remove '/native:0x6AE51D4B' dependency

Jordan2139:
This was causing the script not to run due to a "Native 0x6AE51D4B not supported" error that would basically brick the entire server on the latest recommend FxServer Artifacts


Thekuca:
That is there for a reason. ESX uses a native that is only supported on newer artifacts. You shouldn't use latest recommended, you should use the newest.

CsokiHUN:
As @Thekuca said, it will not be removed.

PR #1049, opened 2023-05-22
Commands & Translations

Cyzxin:
Added /refreshitems To Update The ESX.Items Table

Updated Translation For /refreshjobs


CsokiHUN (approved):

TheFantomas:
Hello,

This pr had to be reverted because a lot of things didn't work afterwards. I would like you to try to fix it and then do it again...

More precisely, the server did not work at all. In multicharacter, the ESX.Jobs table was loaded and therefore the database was marked as not connected and therefore the server could not be turned on.

We can see that in the screenshot.
image
image

Anyway, thanks for the "attempt" to improve the ESX Framework.

With regards and respect
**Legacy Management/Leader** & Developer
**Fantík** & TheKuca

PR #1050, opened 2023-05-22
Create nl_index.html

Arctos2win:


Gellipapa:
@Arctos2win Hi! Thank you for translated index.html but is not good solution.

In the future we will rewrite esx_identity if you like you can do it but then you have to read the translations from Config.lua or JSON and it has to work dynamically.

I don't think it will be accepted because it's not just translated into one language.(NL)

Gellipapa:
@TheFantomas What do you think?

PR #1052, opened 2023-05-25
feat(es_extended): /removeaccountmoney

Cyzxin:


Thekuca (approved):

Tinky124:
Bad, checks are needed here, a person who has 100 dollars with him or in the bank and some admin comes and executes this command, put removeaccountmoney 10000 and the person goes into the minus xD

PR #1058, opened 2023-06-07
refactor(server/player): Optimized player coords updating system

falszywyyy:
The changes I applied to the code regarding self.getCoords(), changing the conditions when joining the server at coordinates to userData.coords and getting rid of the self.updateCoords() function. I believe that my changes when saving are more efficient, because the player's position is not overwritten unnecessarily every second if he changes his position, but the position is saved when leaving the server to the database, and then loaded when entering the server. Have a great day!


bitpredator (approved):

Kenta30 (approved):
Nice

xkariee:
well done

xkariee (approved):
nice!

Thekuca:
Nice PR

Tinky124:
Noo man, but what if server crash or something?

Thekuca:
> Noo man, but what if server crash or something?

It doesn't make a diffrence. In both methods, the database receives the coords on player save not on the updateCoords cycle.

falszywyyy:
> Noo man, but what if server crash or something?

Everything works fine anyway even despite the server crash, I tested this method on a couple of servers where I was a developer and where a large number of people played and every time this method of saving coordinates to the database that I posted here worked smoothly in every case.

Tinky124:
@falszywyyy Oke then

PR #1065, opened 2023-06-12
ox_inventory - startmoney 

TrymTube:
Fixed Starter money and black_money not being given at Char creation


[core]/es_extended/server/main.lua
```diff
@@ -330,6 +330,14 @@ function loadESXPlayer(identifier, playerId, isNew)
...
   else
     exports.ox_inventory:setPlayerInventory(xPlayer, userData.inventory)
+
+    if isNew then
+      for account, money in pairs(Config.StartingAccountMoney) do
+        if account == 'money' or account == 'black_money' then
+          exports.ox_inventory:AddItem(playerId, account, money)
+        end
+      end
+    end
```
Comment (Gellipapa) [inline [core]/es_extended/server/main.lua:340]: If you don't use ox_inventory it will throw an error on the export, so this is wrong. I'll mark where it should be inserted to avoid writing more IFs.

Comment (Gellipapa) [reply [core]/es_extended/server/main.lua:340]: 215 row under.

image

Comment (Gellipapa) [reply [core]/es_extended/server/main.lua:340]: I think if you make the above changes we can accept it afterwards.

Taavi-AU:
could also just add a 

```lua
if GetResourceState("ox_inventory") == "missing" then return end
```

Gellipapa:
> could also just add a
> 
> ```lua
> if GetResourceState("ox_inventory") == "missing" then return end
> ```

Why? We check Config.OxInventory, why need more IF, please check above a picture.

[core]/es_extended/server/main.lua
Comment (TrymTube) [reply [core]/es_extended/server/main.lua:340]: I dont really get what you mean. Because this is basicly already below an else where before if 
`if not Config.OxInventory then` is stated? in the same block also a export for ox inventory for setting the inventory and before that I can not add Items to the players inventory

Screenshot_101

Comment (Gellipapa) [reply [core]/es_extended/server/main.lua:340]: Yes my mistake because github hide the unmodified parts so it's correct how you did it sorry.

Gellipapa (approved):

PR #1073, opened 2023-06-21
Blocking the same character from entering the server

GLN123:


Gellipapa:
@GLN123 Thank you for your work, but we use this when develop server and debugg two client, please add toggle option and toggle key. Config.Debug if false DropPlayer if true return. Thank you.

PR #1078, opened 2023-06-26
refactor: simplify the `GetExtendedPlayers` function

MoskalykA:


Thekuca (approved):

Gellipapa:
@MoskalykA Good job.

If you're going to do it that way, can you make it accept a second parameter table? and return the result as a table.

Here is an example:

local a = GetExtendedPlayers('',{'police','sheriff'})
result
a = {
      sheriff = {SHERIFF_XPLAYERS},
      police = {POLICE_XPLAYERS}
}

and it should continue to work like it used to be an option.

So you have to check if the second parameter is a table and if it is, you have to go through with for and return a table at the end. Thank you very much in advance.

MoskalykA:
> If you're going to do it that way, can you make it accept a second parameter table? and return the result as a table.

I'm doing this in the evening 👍🏻

MoskalykA:
@Gellipapa Can I use 
```lua
{
    ["police"] = true,
    ["pompier"] = true
  }
```
as a second argument, as it seems quicker? https://replit.com/@MoskalykA/esx-GetExtendedPlayers?v=1

Gellipapa:
> 

Thank you, yes but you can create such a list programmatically, but it's perfectly reasonable to understand why indexing is always faster.

Gellipapa (approved):
It looks like a bit of code repetition, but we don't optimize in advance, if there's a problem we'll fix it. Thanks for this feature.

Issue #1089, opened 2023-06-28
[Bug] - Two errors with the commands system


MoskalykA:
**Describe the bug**
If the command doesn't ask for a player but for a vector, for example, and I enter an a, it tells me that the player isn't connected.

A second bug: if I use a command that requires a vector, doing `1 a b` will also give me an error

**To Reproduce**
1. /setcoords a b c -> Player is not online
2. /setcoords 1 b c -> Invalid vector

**Expected behavior**
I should either have the right error or an error

**Screenshots**
/setcoords a b c
![image](https://github.com/esx-framework/esx_core/assets/100430077/f0ba310d-a0fe-4185-bb30-26fca9868b66)
/setcoords 1 b c 
![image](https://github.com/esx-framework/esx_core/assets/100430077/0a2b72aa-4e28-464e-82c6-3f3eacf19c00)

**Debug Info (please complete the following information):**
 - Windows
 - Last
 - Dev

MoskalykA:
@Benzo00 I propose deleting this
![image](https://github.com/esx-framework/esx_core/assets/100430077/2be1cc1e-22c7-46df-acbd-696e0bc7378d)
& and review it
![image](https://github.com/esx-framework/esx_core/assets/100430077/3a66f9d4-acdb-483b-af46-b1e4d8a2799b)

Gellipapa:
@MoskalykA Thank you for report, i will fix today.

Gellipapa:
#1096 Fixed, close issue.

PR #1096, opened 2023-07-07
fix(es_extended/server/functions.lua) register command issue

Gellipapa:
#1089


Gellipapa:
@TheFantomas I fixed 🚀

Issue #1099, opened 2023-07-08
esx_core add new commands


Gellipapa:
Add repair vehicle commands to esx_core.

TheFantomas:
Clean vehicle as well/Add to Repair command SetVehicleDirtLevel native

epyidev:
I have make a pr for that if you want ;)

Gellipapa:
Another issue implement this feature.

PR #1102, opened 2023-07-09
Add repair command

epyidev:
- Add the repair command
- Add missings locales for the repair command (not translated)
- Make french translation for repair command


[core]/es_extended/client/main.lua
```diff
@@ -670,6 +670,28 @@ AddEventHandler("esx:killPlayer", function()
...
+RegisterNetEvent("esx:repairPedVehicle")
+AddEventHandler("esx:repairPedVehicle", function()
+	local GetVehiclePedIsIn = GetVehiclePedIsIn
+
+	ESX.TriggerServerCallback("esx:isUserAdmin", function(admin)
+		if not admin then
+			return
+		end
+		local ped = ESX.PlayerData.ped
+		if IsPedInAnyVehicle(ped, false) then
```
Comment (Gellipapa) [inline [core]/es_extended/client/main.lua:682]: If player not in car use case? Please handle if vehicle nil or 0 players is not a car, and add warning message.

Comment (epyidev) [reply [core]/es_extended/client/main.lua]: It's fixed, let me know if there's anything else :)

[core]/es_extended/server/commands.lua
```diff
@@ -132,6 +132,16 @@ end, false, {
 	}
 })
 
+ESX.RegisterCommand('repair', 'admin', function(xPlayer, args, showError)
+	local ped = GetPlayerPed(xPlayer.source)
+	local pedVehicle = GetVehiclePedIsIn(ped, false)
+	if not pedVehicle or GetPedInVehicleSeat(pedVehicle, -1) ~= ped then
+		return
+		showError(TranslateCap('not_in_vehicle'))
```
Comment (Gellipapa) [inline [core]/es_extended/server/commands.lua:140]: No it's not good, if you return 139 row 140 row not execute. Replace to row 139 to 140 140 to 139.

[core]/es_extended/client/main.lua
```diff
@@ -670,6 +670,24 @@ AddEventHandler("esx:killPlayer", function()
 	SetEntityHealth(ESX.PlayerData.ped, 0)
 end)
 
+RegisterNetEvent("esx:repairPedVehicle")
+AddEventHandler("esx:repairPedVehicle", function()
+	local GetVehiclePedIsIn = GetVehiclePedIsIn
+
```
Comment (Gellipapa) [inline [core]/es_extended/client/main.lua:676]: I check this player in car, if not  add shownotify and return.

Gellipapa:
This is command need support me and id fix. For example: /fix,/repari [playerId] 
And this command prefix use /fix,/repair.

Gellipapa:
#1099

CsokiHUN (approved):
looks good, thanks

Gellipapa:
@epyidev Thank you, it's okey. Please add {fix, repair} command and handle, admin can fix by id. For example: /fix 12 and fix 12 ID player current vehicle.

epyidev:
No problem, I'll do that tomorrow :)

epyidev:
> @epyidev Thank you, it's okey. Please add {fix, repair} command and handle, admin can fix by id. For example: /fix 12 and fix 12 ID player current vehicle.

Added ;)

Gellipapa (changes requested):
Please read my comment and test script.

[core]/es_extended/client/main.lua
```diff
@@ -666,6 +666,17 @@ AddEventHandler("esx:killPlayer", function()
 	SetEntityHealth(ESX.PlayerData.ped, 0)
 end)
 
+RegisterNetEvent("esx:repairPedVehicle")
+AddEventHandler("esx:repairPedVehicle", function()
+	local GetVehiclePedIsIn = GetVehiclePedIsIn
```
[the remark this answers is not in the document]
Comment (epyidev) [reply [core]/es_extended/client/main.lua:671]: Okey, i will do that now ! :)

epyidev:
> Please read my comment and test script.

→ I have removed GetVehiclePedIsIn
→ I have added the discord logging

Link to the test video : https://youtu.be/JlR-0v_oqOw

[core]/es_extended/server/commands.lua
```diff
@@ -132,6 +132,30 @@ end, false, {
...
 
+ESX.RegisterCommand({ 'fix', 'repair' }, 'admin', function(xPlayer, args, showError)
+	local ped = GetPlayerPed(args.playerId.source)
+	local pedVehicle = GetVehiclePedIsIn(ped, false)
+	if not pedVehicle or GetPedInVehicleSeat(pedVehicle, -1) ~= ped then
+		showError(TranslateCap('not_in_vehicle'))
+		return
+	end
+	args.playerId.triggerEvent("esx:repairPedVehicle")
+	xPlayer.showNotification(TranslateCap('command_repair_success'), true, false, 140)
```
Comment (Gellipapa) [inline [core]/es_extended/server/commands.lua:143]: Can you add notify to target. For example admin write command /fix 2. Yes you gave a notify, and target gave notify i think "Admin fix your car" or idk any message.

Comment (Gellipapa) [reply [core]/es_extended/server/commands.lua:143]: and manage the fact that you can't give two notifications. If I have 2 id and I write /fix 2 I don't get two notifications I get one notification.

Gellipapa (changes requested):

[core]/es_extended/server/commands.lua
Comment (epyidev) [reply [core]/es_extended/server/commands.lua:143]: Added !

epyidev:
> Okey, done i will test soon. Thank you for your work.

No problem ! Have a good day !

Gellipapa:
@epyidev Thank you for amazing work  i tested works perfectly i merged.

mahanmoulaei:
The client-side event has no security

Gellipapa:
@mahanmoulaei Hi! How do you think it will be implemented? Would you call him through a server? FreezePlayer, killPlayer are also registered client side events.

mahanmoulaei:
Hey @Gellipapa,

- I suggest passing the vehicle net id through the event to the client and checking if the player's vehicle matches that network id number on client-side. This pretty much increases security. (In very rare cases where the cheater becomes aware of the vehicle net id, this could be abusable. _Again_, in **very rare cases**)

- However, if you really want to step up and make it fully cheat-proof, once this event is triggered you can trigger a server callback to verify if the event call is authorized or not! Yes, this increases the server/client communications, but also increases the security as well!
Something like:

```lua
--- server
local authorizedRepairs = {}

--region /repair command
authorizedRepairs[vehicleEntity] = xPlayer.source

xTarget.triggerEvent("esx:repairPedVehicle")
--region /repair command



--region callback
ESX.RegisterServerCallback("esx:verifyVehicleRepairAuthorization", function(source, cb,)
    local vehicleEntity = GetVehiclePedIsIn(GetPlayerPed(source), false)
    local authorizedIdToRepair = authorizedRepairs[vehicleEntity]
    authorizedRepairs[vehicleEntity] = nil
    
    return cb(authorizedIdToRepair == source)
end)
--region callback
```

```lua
--- client
RegisterNetEvent("esx:repairPedVehicle", function()
    ESX.TriggerServerCallback("esx:verifyVehicleRepairAuthorization", function(authorized)
        if not authorized then return end
        
        local vehicle = GetVehiclePedIsIn(ESX.PlayerData.ped, false)
        SetVehicleEngineHealth(vehicle, 1000)
        SetVehicleEngineOn(vehicle, true, true)
        SetVehicleFixed(vehicle)
        SetVehicleDirtLevel(vehicle, 0)
    end)
end)
```

Gellipapa:
@mahanmoulaei 

Hi! I agree with you, but I think we are a bit overreacting on the security part. 

So, how many cases would you have that would take advantage of that to fix the car. 

Because if we were to rewrite everything and make it safe now, then it would be legitimate for the Legacy to be safe on this feature, but that's not the case at the moment.

If I were a cheater I wouldn't cheat with a fix car command I'd give myself something of value like money, a gun and I think others think the same.

And as you write that more clients and server communication may not be worth it, for us "it's security" because it multiplies the more players there are.

Thanks a lot for the snippet, I'm still thinking about it, but I think it would be more infinity iris to go for security, not necessarily Legacy. 

Also this command was only included because it wasn't and it could be a useful command on a server almost as a basic thing.

PR #1115, opened 2023-07-12
1.0 - Edition (Testing not sure if is all working right!)

C3-FrameWork:
Created and Edited by Its_Cehic.
Why I was making this?
Cause I want that normally every person which using ESX LEGACY using circle progress bar instead defaulted.


esx_circleprogressbar/__resource.lua
```diff
@@ -0,0 +1,16 @@
+resource_manifest_version '44febabe-d386-4d18-afbe-5e627f4af937'
```
Cyzxin, line 1: Why __resource and not fxmanifest? resource is really old and shouldn't be used

Cyzxin (changes requested):

esx_circleprogressbar/__resource.lua
C3-FrameWork: I'm gonna update it!

Cyzxin:
file is fxmanifest not fx_manifest

C3-FrameWork:
an misunderstanding, now it is right!

esx_circleprogressbar/fxmanifest.lua
```diff
@@ -0,0 +1,17 @@
...
+
+author 'Its_Cehic'
+version '1.0'
+description 'Circle Progress Bar - Test Version'
+-- Created and Edited by Its_Cehic
+
+client_script 'script.js'
+server_script 'progress.lua'
+
+ui_page 'index.html'
```
Cyzxin, line 12: these need to be nui/filename not just filename due to file paths. and in the files {} beneath

Cyzxin:
also, i can see a few things that may be issues. make sure this is tested and working, mabye send video clip?

Cyzxin (changes requested):

esx_circleprogressbar/fxmanifest.lua
C3-FrameWork: I tested! Here is error about that is not defined the "document".
Here is the photo:
image

esx_circleprogressbar/nui/script.js
```diff
@@ -0,0 +1,7 @@
+document.addEventListener("DOMContentLoaded", function() {
```
Cyzxin, line 1: that's this line here, what is document and why are you using it? Personally, you should get this working and submit a new PR!

Cyzxin (changes requested):

Gellipapa:
@C3-FrameWork Hi! This is completely wrong, the code is outdated, it is not used correctly. You have a script.js file and it's written in client_script when it has nothing to do with the client. Even if we do it, we'll do it in a way that supports the old one and most likely in some frontend framework so that we can extend it much better in the future and get even better code out of our hands.

C3-FrameWork:
Okay

Gellipapa:
NUI files, you need add files section in fxmanifest not client_script, don't use var in js file.. etc.

PR #1119, opened 2023-07-18
Fixing getNearbyPlayers function source definition

CanysLypys:
By using the function you see the following:

source = GetPlayerPed(source)
-- So source is now the playerPed

source = GetEntityCoords(GetPlayerPed(source))
-- The source is getting new defined by calling the entity coords from the playerPed and then the playerPed once again.

### DIFF
--- a/[core]/es_extended/server/onesync.lua
+++ b/[core]/es_extended/server/onesync.lua
@@ -15,7 +15,7 @@ local function getNearbyPlayers(source, closest, distance, ignore)
 			error("Received invalid first argument (source); should be playerId or vector3 coordinates")
 		end
 
-		source = GetEntityCoords(GetPlayerPed(source))
+		source = GetEntityCoords(source)
 	end
 
 	for _, xPlayer in pairs(ESX.Players) do
--- a/[core]/es_extended/server/onesync.lua
+++ b/[core]/es_extended/server/onesync.lua
@@ -15,12 +15,14 @@ local function getNearbyPlayers(source, closest, distance, ignore)
         
 		if not source then
 			error("Received invalid first argument (source); should be playerId")
+            return result
 		end
 
 		playerCoords = GetEntityCoords(playerPed)
 
         if not playerCoords then
             error("Received nil value (playerCoords); perhaps source is nil at first place?")
+            return result
         end
 	end
 
--- a/[core]/es_extended/server/onesync.lua
+++ b/[core]/es_extended/server/onesync.lua
@@ -38,10 +38,12 @@ local function getNearbyPlayers(source, closest, distance, ignore)
 					result[count] = { id = xPlayer.source, ped = NetworkGetNetworkIdFromEntity(entity), coords = coords, dist = dist }
 				end
 			else
-				local dist = #(playerCoords - coords)
-				if dist <= (result.dist or distance) then
-					result = { id = xPlayer.source, ped = NetworkGetNetworkIdFromEntity(entity), coords = coords, dist = dist }
-				end
+                if xPlayer.source ~= source then
+				    local dist = #(playerCoords - coords)
+				    if dist <= (result.dist or distance) then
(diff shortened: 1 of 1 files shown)


Gellipapa:
@CanysLypys Hi! Thank you for pr. I will need testing after i can merge.

TheFantomas:
> @CanysLypys Hi! Thank you for pr. I will need testing after i can merge.

Tested by me... Its working just fine. Former version its just broken :D Maybe I would do it in a different way.

Maybe I would do it in a different way. On line 12, I would put the source on playerPed and wherever there is a source below, I would put playerPed

TheFantomas:
It will be more clean, because source on line 12 is not source but playerPed...

[core]/es_extended/server/onesync.lua
```diff
@@ -15,7 +15,7 @@ local function getNearbyPlayers(source, closest, distance, ignore)
 			error("Received invalid first argument (source); should be playerId or vector3 coordinates")
 		end
 
-		source = GetEntityCoords(GetPlayerPed(source))
+		source = GetEntityCoords(source)
```
Comment (Gellipapa) [inline [core]/es_extended/server/onesync.lua:18]: It is also logically flawed. So if there is no actual source then GetPlayerPed will be nil and if GetPlayerPed is nil then GetEntityCoords will be nil because GetEntityCoords(nil) is received. 

Here you need to specify that if no source is given then the code should not run. 

The other thing is that it is a good thing to overwrite this variable value, just not very transparent. One minute the source is a player ped id, the next minute it's a coordinate. 

My suggestion is to name the variables correctly, we don't need to waste so much memory and it won't happen like what we want to fix. 

So if ped then local playerPedId,if coordinate then local playerCoords or sourcePedId, sourcePedCoords maybe.

Gellipapa (changes requested):

[core]/es_extended/server/onesync.lua
Comment (TheFantomas) [reply [core]/es_extended/server/onesync.lua:18]: Yea 👍

CanysLypys:
Easy, just wanted to let you know about the problem 👍🏼

[core]/es_extended/server/onesync.lua
```diff
@@ -7,15 +7,21 @@ ESX.OneSync = {}
...
+    local playerCoords
 	if not distance then distance = 100 end
 	if type(source) == 'number' then
-		source = GetPlayerPed(source)
-
+		playerPed = GetPlayerPed(source)
+        
 		if not source then
-			error("Received invalid first argument (source); should be playerId or vector3 coordinates")
+			error("Received invalid first argument (source); should be playerId")
```
Comment (Gellipapa) [inline [core]/es_extended/server/onesync.lua:17]: This is not enough, maybe you should return the result, the currently empty table, to stop the code from running.

[core]/es_extended/server/onesync.lua
```diff
@@ -7,15 +7,21 @@ ESX.OneSync = {}
...
 		end
 
-		source = GetEntityCoords(source)
+		playerCoords = GetEntityCoords(playerPed)
+
+        if not playerCoords then
+            error("Received nil value (playerCoords); perhaps source is nil at first place?")
```
Comment (Gellipapa) [inline [core]/es_extended/server/onesync.lua:23]: This is not enough, maybe you should return the result, the currently empty table, to stop the code from running.

Gellipapa (changes requested):
Please read my messages and fix it. Thank you.

Gellipapa:
@CanysLypys Hi! This is not working, i call function and return me only. 2 people are standing next to each other and yet only me add to table and return. How did you test it?

CanysLypys:
Oop! Haven't forgot to try it with ESX.Players my bad, lmk if the fix of @Thekuca is working, had no time today.

Gellipapa (approved):

Gellipapa:
@CanysLypys @TheFantomas @Thekuca Thank you for your work. I merged.

PR #1123, opened 2023-07-23
Add job2 support to es_extended

nuageeeeee:
I have just finished adding job2 to es_extended so I allow myself to make this pull request which will surely please more than one user.

I added translation to French and English file, sorry for the other language it's the only ones I use (maybe i will try italian)


Mycroft-Studios:
whatever you used to format this, is completely off ngl....

Gellipapa:
@nuageeeeee Hi! Thank you for PR but this is code duplication, if i have 3 jobs, i need made jobs3?  Sorry but this is legacy, we don't support multijobs in legacy. Infinity already knows so it will include.

PR #1124, opened 2023-07-23
Fix setting broken tire

someone005:
It wasn't setting broken tire as broken so it wasnt doing its job

### DIFF
--- a/[core]/es_extended/client/functions.lua
+++ b/[core]/es_extended/client/functions.lua
@@ -1008,9 +1008,7 @@ function ESX.Game.SetVehicleProperties(vehicle, props)
 
     if props.tyreBurst ~= nil then
         for k, v in pairs(props.tyreBurst) do
-            if v then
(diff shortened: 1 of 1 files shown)


Gellipapa:
@someone005 Hi! I don't quite understand what the problem is here. So no v has no value? Because as I see your commit changed so much that you took out the condition and put in v instead of true.

```
        for k, v in pairs(props.tyreBurst) do
            if v then
                SetVehicleTyreBurst(vehicle, tonumber(k), v, 1000.0)
            end
        end
```
So can you try it or can you tell me how you test it because I'd like to see it.

someone005:
When I was testing my garages script I found out that this functions wasn't applying broken tire for some reason. I just removed if statement and replaced true with v and it worked. Sometimes it wasn't working as it should. If this v value would be example false it won't do anything to tire so in my opinion if statement is just useless.

Gellipapa:
I see, it's okey i will test soon. Thank you for PR.

someone005:
No problem ;)

PR #1131, opened 2023-07-26
Update player.lua

katotekii:
fix for `subValue` arg in `setMeta` function

### DIFF
--- a/[core]/es_extended/server/classes/player.lua
+++ b/[core]/es_extended/server/classes/player.lua
@@ -642,7 +642,7 @@ function CreateExtendedPlayer(playerId, identifier, group, accounts, inventory, 
 			if _type ~= "string" then
 				return print(("[^1ERROR^7] xPlayer.setMeta ^5value^7 should be ^5string^7 as a subIndex!"):format(value))
 			end
-
+            self.metadata[index] = table.create(0, 0)
 			self.metadata[index][value] = subValue
 		end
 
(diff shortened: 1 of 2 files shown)


[core]/es_extended/server/classes/player.lua
```diff
@@ -642,7 +642,7 @@ function CreateExtendedPlayer(playerId, identifier, group, accounts, inventory,
 			if _type ~= "string" then
 				return print(("[^1ERROR^7] xPlayer.setMeta ^5value^7 should be ^5string^7 as a subIndex!"):format(value))
 			end
-
+            self.metadata[index] = table.create(0, 0)
```
Comment (Gellipapa) [inline [core]/es_extended/server/classes/player.lua:645]: Why create new table? So it will always be overwritten if there is some data already under the key.

This is not a good solution, what should be done here is to check if metadata[index] exists, if not then create an empty table for it but only if it doesn't exist.

For example: 

```lua
-- data mean index but this is example

if not self.metadata['data'] then
      self.metadata['data'] = {}
end

self.metadata['data']['newKey'] = "any value"

```

Gellipapa (changes requested):
Please make this changes.

katotekii (commented):
yea I realized later, was late here when I posted xd

Gellipapa:
@katotekii Hi! Please update your fork version, and don't commit and push github workflows files.

katotekii:
> @katotekii Hi! Please update your fork version, and don't commit and push github workflows files.

updated

Gellipapa:
@katotekii I fixed metadata sub value. Thank you for open pr.

PR #1134, opened 2023-07-27
Using RPC native for RemoveWeaponComponent

CanysLypys:


Gellipapa:
Hi! @CanysLypys Why commited workflows, please update your forked branch and open new pull request, but server side native okey, but I do not think it is enough that there is a notify. Now if I want to delete from client it was enough to call this event and delete the weapon, now you just do it on server and need to always call xPlayer, not sure if this is the right direction.

ahcenezdh:
You should close this PR + using RPC natives isn't the best thing since some time they are broken

PR #1135, opened 2023-07-28
fix(functions.lua) Change os.time() to os.clock()

epyidev:
os.time() returns the unix time but only seconds which is not very accurate. Better to use os.clock() which returns milliseconds.

### DIFF
--- a/[core]/es_extended/server/functions.lua
+++ b/[core]/es_extended/server/functions.lua
@@ -191,7 +191,7 @@ function Core.SavePlayers(cb)
 		return
 	end
 
-	local startTime <const> = os.time()
+	local startTime <const> = os.clock()
 	local parameters = {}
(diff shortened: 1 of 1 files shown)


Gellipapa:
@epyidev Hi! Thank you for pr and suggestion. 

Explanation:
os.clock() measures the computer's processor time, while os.time() returns the actual time. In this case, we're looking for the real time, how long it took to save the players. Not how much computer processor time it took to save the players. So the fact that we are looking for the real time means that os.time() is implemented and used correctly. 

os.clock() is used where we want to look at the performance of say an algorithm or method, where accuracy is very important but for a player save it doesn't matter, so it's fine if there's a difference between the two, which will be very small if there is, so we stick with the real time, os.time().

PR #1138, opened 2023-07-30
refactor: slight optimisation

MoskalykA:


[core]/es_extended/client/main.lua
```diff
@@ -84,13 +84,13 @@ AddEventHandler('esx:playerLoaded', function(xPlayer, isNew, skin)
 		end)
 	end
 
-	if Config.DisableHealthRegeneration or Config.DisableWeaponWheel or Config.DisableAimAssist or Config.DisableVehicleRewards then
+	if Config.DisableHealthRegeneration then
+		SetPlayerHealthRechargeMultiplier(playerId, 0.0)
```
Comment (Gellipapa) [inline [core]/es_extended/client/main.lua:88]: I don't remember but shouldn't this be handled in a thread?

Comment (MoskalykA) [reply [core]/es_extended/client/main.lua:88]: Why ?

Comment (TheFantomas) [reply [core]/es_extended/client/main.lua:88]: @Thekuca will test that....

Thekuca (approved):
Everything works. Nice PR

PR #1140, opened 2023-08-01
Fix error in chat with admin logs

WirzRexTon:
Error in chat fix --> SCRIPT ERROR : Type 'function' is not supported by json


Cyzxin:
https://github.com/esx-framework/esx_core/blob/dev/%5Bcore%5D/es_extended/server/commands.lua#L33

This is fixed in dev branch. you are using main branch instead, this means it is already fixed for the next ESX update

PR #1146, opened 2023-08-05
Update config.weapons.lua

StrixLuca:
export getWeaponConfig for a MDT or another scripts is very nice
is useful for later or for other scripts

### DIFF
--- a/[core]/es_extended/config.weapons.lua
+++ b/[core]/es_extended/config.weapons.lua
@@ -1007,3 +1007,8 @@ Config.Weapons = {
 		},
 	},
 }
+
+
+exports('getWeaponConfig', function()
+	return Config.Weapons
+end)


Cyzxin:
Could you not just use
`ESX.GetConfig().Weapons`
as the function https://github.com/esx-framework/esx_core/blob/main/%5Bcore%5D/es_extended/common/functions.lua#L23-L25 returns the config variable.

StrixLuca:
> Could you not just use `ESX.GetConfig().Weapons` as the function https://github.com/esx-framework/esx_core/blob/main/%5Bcore%5D/es_extended/common/functions.lua#L23-L25 returns the config variable.

oh lol

Gellipapa:
There is a function for this, so I'll close this pr.

PR #1149, opened 2023-08-07
updated README files

DasistOnly:
updated the documentation urls


Gellipapa:
Thanks for the PR, we will test it soon and if it works we will merge it.

PR #1150, opened 2023-08-07
fix(ESX.UI): Close Opened Menus on Resource Stop

Kenshiin13:
I have encountered several frustrating instances when accidentally restarting a script that still had an open ESX UI. This situation prevents the UI from closing properly and leads to a cascade of errors, ultimately forcing a game restart. To address this issue, I propose the following code changes.

The code introduces an event handler for the onResourceStop event, enabling us to loop through all open ESX UIs and identify any opened menus within the same namespace as the resource name. Following the best practice, the namespace should ideally align with the resource name.

I believe implementing these changes will provide a smoother user experience. I look forward to your feedback and merging these improvements into the main branch.

### DIFF
--- a/[core]/es_extended/client/functions.lua
+++ b/[core]/es_extended/client/functions.lua
@@ -196,6 +196,7 @@ function ESX.UI.Menu.Open(type, namespace, name, data, submit, cancel, change, c
 
     menu.type = type
     menu.namespace = namespace
+    menu.resourceName = (GetInvokingResource() or "Unknown")
     menu.name = name
     menu.data = data
     menu.submit = submit
@@ -1346,7 +1347,7 @@ end)
 AddEventHandler('onResourceStop', function(resourceName)
     for i = 1, #ESX.UI.Menu.Opened, 1 do
         if ESX.UI.Menu.Opened[i] then
-            if ESX.UI.Menu.Opened[i].namespace == resourceName then
+            if ESX.UI.Menu.Opened[i].resourceName == resourceName or ESX.UI.Menu.Opened[i].namespace == resourceName then
                 ESX.UI.Menu.Opened[i].close()
                 ESX.UI.Menu.Opened[i] = nil
             end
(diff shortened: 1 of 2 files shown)


69Gkay:
This problem frustrated me so much it needs to be fixxed asap...

SirAlexS:
it drives me crazy. that NEEDS to be fixed!!

bitpredator (approved):

[core]/es_extended/client/functions.lua
```diff
@@ -1343,6 +1343,17 @@ AddEventHandler('esx:showHelpNotification', function(msg, thisFrame, beep, durat
     ESX.ShowHelpNotification(msg, thisFrame, beep, duration)
 end)
 
+AddEventHandler('onResourceStop', function(resourceName)
+    for i = 1, #ESX.UI.Menu.Opened, 1 do
+        if ESX.UI.Menu.Opened[i] then
```
Comment (Cyzxin) [inline [core]/es_extended/client/functions.lua:1348]: Is this line really needed? you are iterating over only the `ESX.UI.Menu.Opened`, so it will always be a value?

Comment (Kenshiin13) [reply [core]/es_extended/client/functions.lua:1348]:
```lua
function ESX.UI.Menu.Close(type, namespace, name)
    for i = 1, #ESX.UI.Menu.Opened, 1 do
        if ESX.UI.Menu.Opened[i] then
            if ESX.UI.Menu.Opened[i].type == type and ESX.UI.Menu.Opened[i].namespace == namespace and
                ESX.UI.Menu.Opened[i].name == name then
                ESX.UI.Menu.Opened[i].close()
                ESX.UI.Menu.Opened[i] = nil
            end
        end
    end
end
```
I have simply compied the Close function and adapted the conditions. Not sure if it is needed in either cases tbh.

NikartikTheBear:
Your idea is very good as most people use the current resource name as namespace because it's suggested by the docs but if someone for some reason does not use this method, the problem would still arise

Kenshiin13:
> Your idea is very good as most people use the current resource name as namespace because it's suggested by the docs but if someone for some reason does not use this method, the problem would still arise

Ideally you could always include the invoking resource name when opening a menu and just match that instead of the namespace.

Gellipapa:
@Kenshiin13 Hi! Well done, I'll tested and merged. Thanks.

PR #1158, opened 2023-08-17
improve player metadata functions

Arctos2win:


[core]/es_extended/server/classes/player.lua
```diff
@@ -643,15 +643,10 @@ function CreateExtendedPlayer(playerId, identifier, group, accounts, inventory,
 				return print(("[^1ERROR^7] xPlayer.setMeta ^5value^7 should be ^5string^7 as a subIndex!"):format(value))
 			end
 
-            if not self.metadata[index] or type(self.metadata[index]) ~= "table" then
-				self.metadata[index] = { }
-			end
-
+			self.metadata[index] = self.metadata[index] or {}
```
Comment (Gellipapa) [inline [core]/es_extended/server/classes/player.lua:646]: I just merged (1-2 hours), did you test your version? Because I don't think it will work.

Comment (Arctos2win) [reply [core]/es_extended/server/classes/player.lua:646]: i tested the statebagchangehandler part that works bcs its replicated to client, but to be sure i will test it and send a video in staff chat.

[core]/es_extended/server/classes/player.lua
```diff
@@ -643,7 +643,7 @@ function CreateExtendedPlayer(playerId, identifier, group, accounts, inventory,
 				return print(("[^1ERROR^7] xPlayer.setMeta ^5value^7 should be ^5string^7 as a subIndex!"):format(value))
 			end
 
-			self.metadata[index] = self.metadata[index] or {} 
+			self.metadata[index] = type(self.metadata[index]) == 'table' and self.metadata[index] or {}
```
Comment (Gellipapa) [inline [core]/es_extended/server/classes/player.lua:646]: I think you should change the order of this. The first check value is not nil, and the check is after the type of the table. 

Correct version: 
```lua
self.metadata[index] = self.metadata[index] and type(self.metadata[index]) == 'table' or {}
```

Gellipapa (changes requested):
Please read my comment and doing changes.

[core]/es_extended/server/classes/player.lua
Comment (Arctos2win) [reply [core]/es_extended/server/classes/player.lua:646]: nope it returns the value most right so if you have something like this 
local value = statement1 and statement2 and statement3  if statement 1 and 2 are true then the value will 
be set to statement3 thats why the other one needs to come first and since the type of nil is nil and you compare it 
to a string it will return false if the value would be nil

Comment (Kenshiin13) [reply [core]/es_extended/server/classes/player.lua:646]: > I think you should change the order of this. The first check value is not nil, and the check is after the type of the table.
> 
> Correct version:
> 
> ```lua
> self.metadata[index] = self.metadata[index] and type(self.metadata[index]) == 'table' or {}
> ```

this would set the value to true for self.metadata[index] if it is not nil and typeof table.

@Arctos2win version seems correct.

You can re-write that statement using the ternary operator and it will become more clear.
```lua
self.metadata[index] = (type(self.metadata[index]) == 'table' ? self.metadata[index] : {})
```

Comment (Gellipapa) [reply [core]/es_extended/server/classes/player.lua:646]: Yes, right, that's fine. Thanks for the correction.

Gellipapa:
@Arctos2win I can't merge if you add other code. Please open other request. One feature one branch. Thanks.

Issue #1156, opened 2023-08-16
request for CheckDuplicateLicense 


raaazeee:
request for CheckDuplicateLicense like in the qbcore

config.CheckDuplicateLicense if false can 2 account with same identier can join 

if true cant join 2 with the same identier 

for development purposes thanks

Gellipapa:
#1159 Include will this feature request.

PR #1159, opened 2023-08-18
Feature Request 1156: Added CheckDuplicateLicense Config Entry

Kenshiin13:
#### Description
This pull request introduces a new configuration entry, CheckDuplicateLicense, that enhances the identifier management system. When set to true, the system enforces its current behavior by blocking the presence of duplicate identifiers on the server. Conversely, when set to false, the system allows duplicate identifiers for development purposes. The prefix "cl2" is added to the player's identifier, allowing them to connect as a distinct player instance. 

#### Changes Made
- Added a new configuration entry, CheckDuplicateLicense, to the configuration file.
- When CheckDuplicateLicense is set to true, the system retains its current behavior of blocking duplicate identifiers.
- When CheckDuplicateLicense is set to false, duplicate identifiers are allowed on the server. Players are identified by adding the - "cl2" prefix to their identifier, enabling them to connect as a separate player entity.

The code snippet for handling this behaviour has been enhanced as follows:
```lua
function onPlayerJoined(playerId)
    local function handleESXPlayer(identifier)
        local result = MySQL.scalar.await('SELECT 1 FROM users WHERE identifier = ?', {identifier})
        if result then
            loadESXPlayer(identifier, playerId, false)
        else
            createESXPlayer(identifier, playerId)
        end
    end

    local identifier = ESX.GetIdentifier(playerId)
    if identifier then
        if ESX.GetPlayerFromIdentifier(identifier) then
            if (Config.CheckDuplicateLicense) then
                DropPlayer(playerId,
                    ('there was an error loading your character!\nError code: identifier-active-ingame\n\nThis error is caused by a player on this server who has the same identifier as you have. Make sure you are not playing on the same Rockstar account.\n\nYour Rockstar identifier: %s'):format(
                        identifier))
            else
                identifier = string.format("cl2:%s", identifier)
                handleESXPlayer(identifier)
            end
        else
            handleESXPlayer(identifier)
        end
    else
        DropPlayer(playerId,
            'there was an error loading your character!\nError code: identifier-missing-ingame\n\nThe cause of this error is not known, your identifier could not be found. Please come back later or report this problem to the server administration team.')
    end
end
```

#### Reasoning
The introduction of the CheckDuplicateLicense configuration entry offers enhanced control over identifier management. By allowing or disallowing duplicate identifiers based on this setting, the feature caters to both development requirements and server stability. The choice of "cl2" as the prefix is consistent with the convention used by FiveM to permit two concurrent instances.

Thank you for considering this enhancement to the setcoords command. Your feedback and suggestions are appreciated.

[Related GitHub Issue #1156](https://github.com/esx-framework/esx_core/issues/1156)

### DIFF
--- a/[core]/es_extended/config.lua
+++ b/[core]/es_extended/config.lua
@@ -32,7 +32,7 @@ Config.AdminGroups = {
 	['admin'] = true
 }
 
-
+Config.CheckDuplicateLicense     = true      -- Allows duplicate identifiers for development purposes if set to false. Should be set to true for production!
 Config.EnablePaycheck            = true      -- enable paycheck
 Config.LogPaycheck               = false     -- Logs paychecks to a nominated Discord channel via webhook (default is false)
 Config.EnableSocietyPayouts      = false     -- pay from the society account that the player is employed at? Requirement: esx_society
--- a/[core]/es_extended/server/main.lua
+++ b/[core]/es_extended/server/main.lua
@@ -45,24 +45,33 @@ else
 end
 
 function onPlayerJoined(playerId)
-	local identifier = ESX.GetIdentifier(playerId)
-	if identifier then
-		if ESX.GetPlayerFromIdentifier(identifier) then
-			DropPlayer(playerId,
-				('there was an error loading your character!\nError code: identifier-active-ingame\n\nThis error is caused by a player on this server who has the same identifier as you have. Make sure you are not playing on the same Rockstar account.\n\nYour Rockstar identifier: %s'):format(
-					identifier))
-		else
-			local result = MySQL.scalar.await('SELECT 1 FROM users WHERE identifier = ?', { identifier })
-			if result then
-				loadESXPlayer(identifier, playerId, false)
-			else
-				createESXPlayer(identifier, playerId)
-			end
-		end
-	else
-		DropPlayer(playerId,
-			'there was an error loading your character!\nError code: identifier-missing-ingame\n\nThe cause of this error is not known, your identifier could not be found. Please come back later or report this problem to the server administration team.')
-	end
+    local function handleESXPlayer(identifier)
+        local result = MySQL.scalar.await('SELECT 1 FROM users WHERE identifier = ?', {identifier})
+        if result then
+            loadESXPlayer(identifier, playerId, false)
+        else
+            createESXPlayer(identifier, playerId)
+        end
+    end
+
+    local identifier = ESX.GetIdentifier(playerId)
+    if identifier then
+        if ESX.GetPlayerFromIdentifier(identifier) then
+            if (Config.CheckDuplicateLicense) then
+                DropPlayer(playerId,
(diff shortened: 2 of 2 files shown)


CsokiHUN:
We also need to add a check in the `esx_multicharacter` script

https://github.com/esx-framework/esx_core/blob/a95e0397b05417c9d42126268103d2682fdb17ba/%5Bcore%5D/esx_multicharacter/server/main.lua#L120

Kenshiin13:
> We also need to add a check in the `esx_multicharacter` script
> 
> 
> 
> https://github.com/esx-framework/esx_core/blob/a95e0397b05417c9d42126268103d2682fdb17ba/%5Bcore%5D/esx_multicharacter/server/main.lua#L120

You're right, I forgot about multichar. Will push a commit to fix this later.

Gellipapa:
@Kenshiin13 Hi! Can you fix that?

Kenshiin13:
> ```lua
> Config.CheckDuplicateLicense
> ```

On it.

Kenshiin13:
> I don't quite understand, but if you meant that if Debug is true then it's okey.


I will push a commit to support multicharacter soon is what I meant.

Kenshiin13:
Still debating on a good way to do this for multi char. No as straightforward, unless I'm missing something. Got an idea? @Gellipapa

vrnagy:
You could use convar for it like `Config.CheckDuplicateLicense = GetConvar('esx:cl2', 'false') == 'false'` then use it in the core and the multichar resource. If somebody wants to enable the cl2 flag then simply set the convar to true.

Kenshiin13:
> You could use convar for it like `Config.CheckDuplicateLicense = GetConvar('esx:cl2', 'false') == 'false'` then use it in the core and the multichar resource. If somebody wants to enable the cl2 flag then simply set the convar to true.

That part is already done. I personally hate working with convars, so I'd rather not do that. I am debating on how to assign a player the "cl2" prefix for multichar in an efficient way. Because, you know, the first client can have any prefix "char1:", "char2:" etc.

Gellipapa:
@Kenshiin13 Hi! What do you know about FiveM fixing the two client connection? 

In what other cases could this be used? Because if it's for development purposes, it's not worth bothering with it because you can't connect 2 clients at the same time as far as I know.

In what other use cases can this be used?

vrnagy:
With the proposed PR You can connect to the server with the secondary client.

Gellipapa:
> With the proposed PR You can connect to the server with the secondary client.

Yes, that's understandable, but I understand that the FiveM level doesn't allow two client connections anymore, so the framework is completely independent.

vrnagy:
The cl2 flag still works and I can't find anything related on the fivem github that indicates it will be removed.

Gellipapa:
> The cl2 flag still works and I can't find anything related on the fivem github that indicates it will be removed.

Thanks for the feedback, will try it soon.

Gellipapa:
Yes I checked it still works to connect the two clients on the FiveM side, so it might be worth managing it in ESX.

Gellipapa:
I did some research and there is currently an option within es_extended called Debug that handles the 2 client connections. 

What was the problem with this or why do I need a new logic? 

Thanks for the reply in advance.

Kenshiin13:
Didnt know that. PR can be closed!

PR #1161, opened 2023-08-20
Security Enhancement: Distance Check for 'esx:onPickup' Event

Kenshiin13:
#### Description

This pull request introduces a critical security update for the "esx:onPickup" event to mitigate potential abuses. The update incorporates a distance check mechanism, preventing players from exploiting the event to collect pickups from excessive distances. This measure ensures fair gameplay and maintains the integrity of the event.

#### Changes Made

- Store serverside pickup coords as a vector, eliminating the need for client-side conversion, and allowing for arithmetic operations on them. 
- Implemented a distance check within the existing 'esx:onPickup' event handling code.
- If a player attempts to interact with a pickup beyond a specified distance (5.0 units), a warning is triggered, and the interaction is blocked.

The modified code snippet is as follows:
```lua
RegisterNetEvent('esx:onPickup')
AddEventHandler('esx:onPickup', function(pickupId)
    local pickup, xPlayer, success = Core.Pickups[pickupId], ESX.GetPlayerFromId(source)

    if pickup then
        local playerPickupDistance = #(pickup.coords - xPlayer.getCoords(true))
        if(playerPickupDistance > 5.0) then
            print(('[^3WARNING^7] Player Detected Cheating (Out of range pickup): ^5%s^7'):format(xPlayer.getIdentifier()))
            return
        end

        -- Rest of the original event handling code

        if success then
            Core.Pickups[pickupId] = nil
            TriggerClientEvent('esx:removePickup', -1, pickupId)
        end
    end
end)
```

#### Reasoning

This security update is vital to prevent potential abuses of the "esx:onPickup" event. The event could potentially be exploited by malicious players to trigger unauthorized pickups using a simple code snippet like this:
```lua
RegisterNetEvent("esx:createPickup", function(pickupId)
    TriggerServerEvent("esx:onPickup", pickupId)
end)
 ```
By implementing a distance check, the update ensures that players can only interact with pickups within a reasonable range, discouraging unauthorized collection of pickups from afar.

#### Counterargument Consideration:

It's worth addressing the potential counterargument that the distance check might be unnecessary due to a player's ability to teleport to pickups, thus eliminating the need for this change. However, it's essential to recognize that third-party anticheats are designed precisely to identify players employing teleportation tactics across the map. Additionally, the presence of one form of vulnerability should not overshadow addressing another potential attack vector.

Thank you for considering these enhancements. Your feedback and suggestions are appreciated.

### DIFF
--- a/[core]/es_extended/server/functions.lua
+++ b/[core]/es_extended/server/functions.lua
@@ -469,19 +469,19 @@ function ESX.GetUsableItems()
 end
 
 if not Config.OxInventory then
-	function ESX.CreatePickup(type, name, count, label, playerId, components, tintIndex, coords)
+	function ESX.CreatePickup(itemType, name, count, label, playerId, components, tintIndex, coords)
 		local pickupId = (Core.PickupId == 65635 and 0 or Core.PickupId + 1)
 		local xPlayer = ESX.Players[playerId]
 		coords = ( (type(coords) == "vector3" or type(coords) == "vector4") and coords.xyz or xPlayer.getCoords(true))
 
-		Core.Pickups[pickupId] = { type = type, name = name, count = count, label = label, coords = coords }
+		Core.Pickups[pickupId] = { type = itemType, name = name, count = count, label = label, coords = coords }
 
-		if type == 'item_weapon' then
+		if itemType == 'item_weapon' then
 			Core.Pickups[pickupId].components = components
 			Core.Pickups[pickupId].tintIndex = tintIndex
 		end
 
-		TriggerClientEvent('esx:createPickup', -1, pickupId, label, coords, type, name, components, tintIndex)
+		TriggerClientEvent('esx:createPickup', -1, pickupId, label, coords, itemType, name, components, tintIndex)
 		Core.PickupId = pickupId
 	end
 end
--- a/[core]/es_extended/server/functions.lua
+++ b/[core]/es_extended/server/functions.lua
@@ -472,7 +472,7 @@ if not Config.OxInventory then
 	function ESX.CreatePickup(type, name, count, label, playerId, components, tintIndex)
 		local pickupId = (Core.PickupId == 65635 and 0 or Core.PickupId + 1)
 		local xPlayer = ESX.Players[playerId]
-		local coords = xPlayer.getCoords()
+		local coords = xPlayer.getCoords(true)
 
 		Core.Pickups[pickupId] = { type = type, name = name, count = count, label = label, coords = coords }
 
--- a/[core]/es_extended/server/functions.lua
+++ b/[core]/es_extended/server/functions.lua
@@ -469,10 +469,10 @@ function ESX.GetUsableItems()
 end
 
 if not Config.OxInventory then
-	function ESX.CreatePickup(type, name, count, label, playerId, components, tintIndex)
+	function ESX.CreatePickup(type, name, count, label, playerId, components, tintIndex, coords)
 		local pickupId = (Core.PickupId == 65635 and 0 or Core.PickupId + 1)
 		local xPlayer = ESX.Players[playerId]
-		local coords = xPlayer.getCoords(true)
+		coords = ( (type(coords) == "vector3" or type(coords) == "vector4") and coords.xyz or xPlayer.getCoords(true))
 
 		Core.Pickups[pickupId] = { type = type, name = name, count = count, label = label, coords = coords }
 
--- a/[core]/es_extended/client/functions.lua
+++ b/[core]/es_extended/client/functions.lua
@@ -70,17 +70,17 @@ function ESX.Progressbar(message, length, Options)
     print("[^1ERROR^7] ^5ESX Progressbar^7 is Missing!")
 end
 
-function ESX.ShowNotification(message, type, length)
+function ESX.ShowNotification(message, notifyType, length)
     if GetResourceState("esx_notify") ~= "missing" then
-        return exports["esx_notify"]:Notify(type, length, message)
+        return exports["esx_notify"]:Notify(notifyType, length, message)
     end
 
     print("[^1ERROR^7] ^5ESX Notify^7 is Missing!")
 end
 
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Review by iSentrie — CHANGES_REQUESTED** [review]
this will require decrease in vector z, else player won't be able to pick anything from ground

```lua
TriggerEvent('esx:createPickup', pickupId, pickup.label, pickup.coords - vector3(0, 0, 0.5), pickup.type, pickup.name, pickup.components, pickup.tintIndex)
```

however, who don't use ox_inventory, after client restart missing pickups will appear floating in the air, needs an better fix, but for now at least it will work. currently in last version missing pickups wasn't working at all.

Kenshiin13:
> this will require decrease in vector z, else player won't be able to pick anything from ground

Haven't tested that tbh. The pickups spawned properly on the ground when I tested the vulnerability. I can push another commit tho if requested.

iSentrie:
i tested, that's why i wrote it

Kenshiin13:
> your changes fixes the error

My changes have nothing to do with that.

iSentrie:
> > your changes fixes the error
> 
> My changes have nothing to do with that.

without your changes missing pickups won't spawn at all, resulting in an error in client console.
you fixed it without knowing it.
it was fixed before by someone and then broken again by someone, now you fixed it again.

Gellipapa:
@Kenshiin13 Hi! Could you write in to accept coordinates not just the current coordinate? 
Example: I need to tell you that I want a pickup points xyz, because currently it only looks at your current coordinate to the best of my knowledge.

Kenshiin13:
> @Kenshiin13 Hi! Could you write in to accept coordinates not just the current coordinate? Example: I need to tell you that I want a pickup points xyz, because currently it only looks at your current coordinate to the best of my knowledge.

I'm sorry, I'm not sure what you mean. Why would we test against anything other than the players current coordinates to determine whether a pickup request ist legitmate or not?

Gellipapa:
> 
I seem to remember that currently in ESX there is no option to spawn a pickup point to a given coordinate, you can only spawn a point to the player's coordinate. That's what I was thinking about, that I should be able to spawn a pickup point to a custom coordinate.

Kenshiin13:
> > 
> 
> I seem to remember that currently in ESX there is no option to spawn a pickup point to a given coordinate, you can only spawn a point to the player's coordinate. That's what I was thinking about, that I should be able to spawn a pickup point to a custom coordinate.

You mean like this 92d6f49464fbffc2f5431c06f560637eb457302a?

Gellipapa:
@Kenshiin13 Yes, thank you.

Gellipapa:
Thanks for the PR, we will test it soon and if it works we will merge it.

Gellipapa:
@Kenshiin13 Hi! I tested this pr and i found bug.

- ESX.CreatePickup method first parameter is type (this is not correct name, please fix) and this is name overwrite default lua type checker and i get nil value if i try run this code. (Please rename all type parameter to other name in server/main.lua)

Let's find a solution so that if I don't want to spawn a weapon, I don't have to specify "components" and "tintIdex", because I currently have to specify them because coordinates are included and the parameter number of pieces makes it mandatory to specify them.  **(!!We have to pay attention to backwards compatibility because it is not easy to do this.)**


image

[core]/es_extended/server/functions.lua
```diff
@@ -469,10 +469,10 @@ function ESX.GetUsableItems()
 end
 
 if not Config.OxInventory then
-	function ESX.CreatePickup(type, name, count, label, playerId, components, tintIndex)
+	function ESX.CreatePickup(type, name, count, label, playerId, components, tintIndex, coords)
```
Comment (Gellipapa) [inline [core]/es_extended/server/functions.lua:472]: Rename type parameter because overwrite default lua type checker method.

Gellipapa (changes requested):

Gellipapa:
In the future, please make sure that it is tested, because this is a basic bug that could have been fixed in the first test. Thanks.

Kenshiin13:
> Let's find a solution so that if I don't want to spawn a weapon, I don't have to specify "components" and "tintIdex", because I currently have to specify them because coordinates are included and the parameter number of pieces makes it mandatory to specify them. **(!!We have to pay attention to backwards compatibility because it is not easy to do this.)**
> 
> image

Well, I am not sure if there is a way to do this without breaking backwards-compatibility.

Gellipapa:
> > Let's find a solution so that if I don't want to spawn a weapon, I don't have to specify "components" and "tintIdex", because I currently have to specify them because coordinates are included and the parameter number of pieces makes it mandatory to specify them. **(!!We have to pay attention to backwards compatibility because it is not easy to do this.)**
> 
> Well, I am not sure if there is a way to do this without breaking backwards-compatibility.

If you have any ideas, feel free to write them down, I can't think of anything right now, if we don't have a solution, it will stay as it was.

Kenshiin13:
> If you have any ideas, feel free to write them down, I can't think of anything right now, if we don't have a solution, it will stay as it was. I mean that part will not be changed.

Nope, not really. Does that work for you? https://github.com/esx-framework/esx_core/pull/1161/commits/1b27475ce16abbee6aeec606f9bf123069b5f081

[core]/es_extended/server/functions.lua
Comment (Kenshiin13) [reply [core]/es_extended/server/functions.lua]: https://github.com/esx-framework/esx_core/pull/1161/commits/1b27475ce16abbee6aeec606f9bf123069b5f081

Gellipapa:
> > If you have any ideas, feel free to write them down, I can't think of anything right now, if we don't have a solution, it will stay as it was. I mean that part will not be changed.
> 
> Nope, not really. Does that work for you? [1b27475](https://github.com/esx-framework/esx_core/commit/1b27475ce16abbee6aeec606f9bf123069b5f081)

I will check soon.

Gellipapa:
Hi! Well done, I was tested and merged.

Gellipapa (approved):

PR #1165, opened 2023-08-20
refactor: disable action events

MoskalykA:


Gellipapa:
Hi! @MoskalykA  Why would it make sense to turn off something that helps development and some scripts are based on it? like esx_hud, esx_cruisecontrol etc.

MoskalykA:
> Hi! @MoskalykA Why would it make sense to turn off something that helps development and some scripts are based on it? like esx_hud, esx_cruisecontrol etc.

That's why it's possible to deactivate it. Personally, I don't use it and would like to have as few operations as possible.

Gellipapa:
> > Hi! @MoskalykA Why would it make sense to turn off something that helps development and some scripts are based on it? like esx_hud, esx_cruisecontrol etc.
> 
> That's why it's possible to deactivate it. Personally, I don't use it and would like to have as few operations as possible.

What disadvantage do you have if you don't use it and it's not switched off? 

We have to think of the majority here, if there is no real world example that makes it useful we will not accept it, because it can generate an error if someone turns it off and later starts using, say, esx_hud.

MoskalykA:
> What disadvantage do you have if you don't use it and it's not switched off?
> 
> We have to think of the majority here, if there is no real world example that makes it useful we will not accept it, because it can generate an error if someone turns it off and later starts using, say, esx_hud.

There's no disadvantage if you use this system, but if you don't then you might as well disable it.

"because it can generate an error if someone turns it off and later starts using"
This could be the same with any similar changes, why did you do it with `getSharedObject` then?

Gellipapa:
> > What disadvantage do you have if you don't use it and it's not switched off?
> > We have to think of the majority here, if there is no real world example that makes it useful we will not accept it, because it can generate an error if someone turns it off and later starts using, say, esx_hud.
> 
> There's no disadvantage if you use this system, but if you don't then you might as well disable it.
> 
> "because it can generate an error if someone turns it off and later starts using" This could be the same with any similar changes, why did you do it with `getSharedObject` then?

I didn't do that, that was another development team, but now I'm not going to bullshit people into putting it back, I hope you understand thanks.

Gellipapa:
Of course, there's only one level after which it's not worth configuring, because then everything that's not used by someone could be turned off. That's why the getSharedObject should not be enabled/disabled because everyone would set it to true and use the old method.

These addeventhandlers don't break performance, because they run when someone triggers them.

MoskalykA:
How do we detect their triggers?

Gellipapa:
> 

What do you mean? You can see example in esx_hud, or esx_cruisecontrol.

MoskalykA:
> What do you mean? You can see example in esx_hud, or esx_cruisecontrol.

I wanted you to point out the triggering of these events but oh well. Personally, having an iteration every 200 milliseconds for a system I don't use bothers me.

Kenshiin13:
> > What do you mean? You can see example in esx_hud, or esx_cruisecontrol.
> 
> I wanted you to point out the triggering of these events but oh well. Personally, having an iteration every 200 milliseconds for a system I don't use bothers me.

Script developers may rely on these events, could be kind of annoying if the user turns it off and wonders why his script does not work.

MoskalykA:
> Script developers may rely on these events, could be kind of annoying if the user turns it off and wonders why his script does not work.

But why assume that a person would deactivate something he doesn't know the meaning of, and then even if he did, that person would be helped?

Kenshiin13:
> But why assume that a person would deactivate something he doesn't know the meaning of, and then even if he did, that person would be helped?

People turn stuff off they don't know about all the time. 
I'm not necessarily against the changes, just making sure we consider all the outcomes.

Gellipapa:
@MoskalykA 

We will not accept it because there is no use case where it would be useful to introduce it. 99% of people would leave it as it is because it would be the default.(true) If it bothers you then comment it out in your own system or implement it yourself. 

Thank you for your understanding.

Issue #1151, opened 2023-08-09
GetExtendedPlayers refactor


Gellipapa:
We need to modify the function a bit, because it is not the best to use in some cases.

Let's look at an example: 
I would like to know how many police and sheriffs there are in total.

local policesheriffCount = ESX.GetExtendedPlayers('job', {'police', 'sheriff'})
The result is: "sheriff:{}, police:{{xplayer},{xplayer}}"
Now to find out how many police and sheriffs there are in total. I have to double for loop because I get two table back.

Look something like this:
```
        local datas = ESX.GetExtendedPlayers('job', {'police', 'sheriff'})
        local count = 0
        for _,dataValue in pairs(datas) do
            for _, v in pairs(dataValue) do
                count = count + 1
            end
        end
```
A solution to this could be that if the third parameter is true, then return everything in a table or if the third parameter is 'count' then just count them or something like that.

The point is to make it as easy to use as possible and dynamic enough. 

In most cases 98% of the scripts are used for count, the rest for xPlayer. So there are not that many cases but you have to find a solution because double for loop is not a solution.

CsokiHUN:
I think a separate function would be better, like ESX.GetNumOfPlayers({'police', 'serif'})

Anyway, good idea, just need a good implementation so that the server doesn't flame if there are too many players

TrymTube:
if you will make that a seperate function only one will be used in the future or one gets removed ( in this case the one that does less stuff )
I find it a pretty simple solution with the table in GetExtendedPlayers

Gellipapa:
> if you will make that a seperate function only one will be used in the future or one gets removed ( in this case the one that does less stuff ) I find it a pretty simple solution with the table in GetExtendedPlayers

Hi! Can you explain what you mean by "simple solution"?
This is currently also manage a table, not this problem.

Kenshiin13:
> if you will make that a seperate function only one will be used in the future or one gets removed ( in this case the one that does less stuff ) I find it a pretty simple solution with the table in GetExtendedPlayers

These two functions serve distinct purposes. One function provides a complete handle to the player object, while the other function only returns the count of players that meet specific criteria. Introducing an additional function won't render the other one useless; instead, it will provide a more distinct and defined role for each.

thelindat:
> I have to double for loop because I get two table back.

Really questioning the choices constantly being made on this project.. Either way it returns array-like tables, so you don't need any loops.

[...]

PR #1169, opened 2023-08-20
Feature Request 1151: Added ESX.GetNumPlayers Function

Kenshiin13:
#### Description
This pull request introduces the [requested ](https://github.com/esx-framework/esx_core/issues/1151) ESX.GetNumPlayers function, providing improved usability and performance when retrieving player counts based on specific key-value pairs. The new function streamlines the process of counting players associated with particular attributes, eliminating the need for a double loop and offering enhanced intuitiveness in usage.

#### Changes Made
- The function ESX.GetNumPlayers has been introduced

The code snippet for the function is as follows:
```lua
function ESX.GetNumPlayers(key, val)
    if(not key) then
        return ESX.Table.SizeOf(ESX.Players)
    end

    local isValATable = (type(val) == "table")
    local valTable = (isValATable and val or {val})
    local numPlayers = {}
    for i, val in ipairs(valTable) do
        numPlayers[val] = 0
    end    

    for i, xPlayer in pairs(ESX.Players) do
        local value = (key == "job" and xPlayer.job.name or xPlayer[key])
        if(numPlayers[value]) then
            numPlayers[value] += 1
        end 
    end

    return (isValATable and numPlayers or numPlayers[val])
end
```

#### Reasoning
The new ESX.GetNumPlayers function offers an efficient and straightforward approach to obtaining player counts based on specific attributes. By eliminating the need for a double loop, this enhancement contributes to better performance and code readability. The update also aligns with the goal of creating a more intuitive experience for developers utilizing the function.

Thank you for considering this enhancement. Your feedback and suggestions are highly valued.

[Related GitHub Issue #1151](https://github.com/esx-framework/esx_core/issues/1151)

### DIFF
--- a/[core]/es_extended/server/functions.lua
+++ b/[core]/es_extended/server/functions.lua
@@ -261,24 +261,31 @@ end
 
 function ESX.GetNumPlayers(key, val)
     if(not key) then
-        return ESX.Table.SizeOf(ESX.Players)
+        return #ESX.GetPlayers()
     end
 
-    local isValATable = (type(val) == "table")
-    local valTable = (isValATable and val or {val})
-    local numPlayers = {}
-    for i, val in ipairs(valTable) do
-        numPlayers[val] = 0
-    end    
-
-    for i, xPlayer in pairs(ESX.Players) do
-        local value = (key == "job" and xPlayer.job.name or xPlayer[key])
-        if(numPlayers[value]) then
-            numPlayers[value] += 1
-        end 
+    if type(val) == "table" then
+        local numPlayers = {}
+        if key == "job" then
+            for _, v in ipairs(val) do
+                numPlayers[v] = (ESX.JobsPlayerCount[v] or 0)
+            end
+            return numPlayers
+        else
+            local filteredPlayers = ESX.GetExtendedPlayers(key, val)
+            for i, v in pairs(filteredPlayers) do
+                numPlayers[i] = (#v or 0)
+            end
+            return numPlayers
+        end
+    else
+        if key == "job" then
+            return (ESX.JobsPlayerCount[val] or 0)
+        else
+            local filteredPlayers = ESX.GetExtendedPlayers(key, val)
+            return #filteredPlayers
+        end
     end
-
-    return (isValATable and numPlayers or numPlayers[val])
 end
 
 function ESX.GetPlayerFromId(source)
--- a/[core]/es_extended/server/functions.lua
+++ b/[core]/es_extended/server/functions.lua
@@ -260,7 +260,7 @@ function ESX.GetExtendedPlayers(key, val)
 end
 
 function ESX.GetNumPlayers(key, val)
-    if(not key) then
+    if not key then
         return #ESX.GetPlayers()
     end
 
--- a/[core]/es_extended/server/functions.lua
+++ b/[core]/es_extended/server/functions.lua
@@ -264,7 +264,8 @@ function ESX.GetNumPlayers(key, val)
         return ESX.Table.SizeOf(ESX.Players)
     end
 
-    local valTable = (type(val) == "table" and val or {val})
+    local isValATable = (type(val) == "table")
+    local valTable = (isValATable and val or {val})
     local numPlayers = {}
     for i, val in ipairs(valTable) do
         numPlayers[val] = 0
@@ -277,7 +278,7 @@ function ESX.GetNumPlayers(key, val)
         end 
     end
 
-    return numPlayers
+    return (isValATable and numPlayers or numPlayers[val])
 end
 
 function ESX.GetPlayerFromId(source)
--- a/[core]/es_extended/server/functions.lua
+++ b/[core]/es_extended/server/functions.lua
@@ -284,9 +284,7 @@ function ESX.GetNumPlayers(key, val)
         return (ESX.JobsPlayerCount[val] or 0)
     end
 
-    local filteredPlayers = ESX.GetExtendedPlayers(key, val)
-    return #filteredPlayers
-    end
+    return #ESX.GetExtendedPlayers(key, val)
 end
 
 function ESX.GetPlayerFromId(source)
--- a/[core]/es_extended/server/functions.lua
+++ b/[core]/es_extended/server/functions.lua
@@ -271,20 +271,21 @@ function ESX.GetNumPlayers(key, val)
                 numPlayers[v] = (ESX.JobsPlayerCount[v] or 0)
             end
             return numPlayers
-        else
-            local filteredPlayers = ESX.GetExtendedPlayers(key, val)
-            for i, v in pairs(filteredPlayers) do
-                numPlayers[i] = (#v or 0)
-            end
-            return numPlayers
         end
-    else
-        if key == "job" then
-            return (ESX.JobsPlayerCount[val] or 0)
-        else
-            local filteredPlayers = ESX.GetExtendedPlayers(key, val)
-            return #filteredPlayers
+
+        local filteredPlayers = ESX.GetExtendedPlayers(key, val)
+        for i, v in pairs(filteredPlayers) do
+            numPlayers[i] = (#v or 0)
         end
+        return numPlayers
+    end
+
+    if key == "job" then
+        return (ESX.JobsPlayerCount[val] or 0)
+    end
+
+    local filteredPlayers = ESX.GetExtendedPlayers(key, val)
+    return #filteredPlayers
     end
 end
 
--- a/[core]/es_extended/server/functions.lua
+++ b/[core]/es_extended/server/functions.lua
@@ -259,6 +259,27 @@ function ESX.GetExtendedPlayers(key, val)
 	return xPlayers
 end
 
+function ESX.GetNumPlayers(key, val)
+    if(not key) then
+        return ESX.Table.SizeOf(ESX.Players)
+    end
+
+    local valTable = (type(val) == "table" and val or {val})
+    local numPlayers = {}
+    for i, val in ipairs(valTable) do
+        numPlayers[val] = 0
+    end    
+
+    for i, xPlayer in pairs(ESX.Players) do
+        local value = (key == "job" and xPlayer.job.name or xPlayer[key])
+        if(numPlayers[value]) then
+            numPlayers[value] += 1
+        end 
+    end
+
+    return numPlayers
+end
+
 function ESX.GetPlayerFromId(source)
 	return ESX.Players[tonumber(source)]
 end
--- a/[core]/es_extended/server/main.lua
+++ b/[core]/es_extended/server/main.lua
@@ -378,7 +378,7 @@ AddEventHandler('playerDropped', function(reason)
 	if xPlayer then
 		TriggerEvent('esx:playerDropped', playerId, reason)
         local job = xPlayer.getJob().name
-        ESX.JobsPlayerCount[job] = (ESX.JobsPlayerCount[job] or 1) -1
+        ESX.JobsPlayerCount[job] = ((ESX.JobsPlayerCount[job] and ESX.JobsPlayerCount[job] > 0) and ESX.JobsPlayerCount[job] or 1) -1
 
 		Core.playersByIdentifier[xPlayer.identifier] = nil
 		Core.SavePlayer(xPlayer, function()
@@ -393,7 +393,7 @@ AddEventHandler("esx:playerLoaded", function(playerId, xPlayer, isNew)
 end)
 
 AddEventHandler("esx:setJob", function(src, job, lastJob)
-    ESX.JobsPlayerCount[lastJob.name] = (ESX.JobsPlayerCount[lastJob.name] or 1) - 1
+    ESX.JobsPlayerCount[lastJob.name] = ((ESX.JobsPlayerCount[lastJob.name] and ESX.JobsPlayerCount[lastJob.name] > 0) and ESX.JobsPlayerCount[lastJob.name] or 1) -1
     ESX.JobsPlayerCount[job.name] = (ESX.JobsPlayerCount[job.name] or 0) + 1
 end)
 
--- a/[core]/es_extended/server/main.lua
+++ b/[core]/es_extended/server/main.lua
@@ -377,12 +377,24 @@ AddEventHandler('playerDropped', function(reason)
 
 	if xPlayer then
 		TriggerEvent('esx:playerDropped', playerId, reason)
+        local job = xPlayer.getJob().name
+        ESX.JobsPlayerCount[job] = (ESX.JobsPlayerCount[job] or 1) -1
 
 		Core.playersByIdentifier[xPlayer.identifier] = nil
 		Core.SavePlayer(xPlayer, function()
 			ESX.Players[playerId] = nil
 		end)
 	end
+end)
+
+AddEventHandler("esx:playerLoaded", function(playerId, xPlayer, isNew)
+    local job = xPlayer.getJob().name
+    ESX.JobsPlayerCount[job] = (ESX.JobsPlayerCount[job] or 0) +1
+end)
+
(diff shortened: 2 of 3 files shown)


[core]/es_extended/server/functions.lua
```diff
@@ -259,6 +259,27 @@ function ESX.GetExtendedPlayers(key, val)
 	return xPlayers
 end
 
+function ESX.GetNumPlayers(key, val)
+    if(not key) then
+        return ESX.Table.SizeOf(ESX.Players)
```
Comment (Cyzxin) [inline [core]/es_extended/server/functions.lua:264]: why not just `#ESX.Players` most likely more efficient correct me if i'm wrong

Comment (Kenshiin13) [reply [core]/es_extended/server/functions.lua:264]: The table ESX.Players contains nil value keys (most noteably when players drop from the server), so we have to use the nil proof alternative `ESX.Table.SizeOf(t)`

Comment (Cyzxin) [reply [core]/es_extended/server/functions.lua:264]: Odd. I would've expected them to be removed on disconnect. probably there for some reason i'm unsure of but yeah I get what you mean now, apologies.

Kenshiin13:
I don't really like that the function returns a table even if the function param "val" is not a table. It should return a number directly, that is much more intuitive. Let me fix that real quick..

Arctos2win:
> 
just get the table length of GetExtendedPlayers function not that hard

Kenshiin13:
> > 
> 
> just get the table length of GetExtendedPlayers function not that hard

GetExtendedPlayers  returns ESX.Players, so we are left with the same issue, aren't we?

Arctos2win:
> > > 
> > 
> > 
> > just get the table length of GetExtendedPlayers function not that hard
> 
> GetExtendedPlayers returns ESX.Players, so we are left with the same issue, aren't we?

well as i mentioned before this is interesting if we talk about job / group counting
but otherwise you could just get ESX.Table.SizeOf(t)

Kenshiin13:
> well as i mentioned before this is interesting if we talk about job / group counting but otherwise you could just get ESX.Table.SizeOf(t)

Also, you would have to trust developers to know that ESX.Players contains nil values and they can't just do `#ESX.GetExtendedPlayers()`.

CsokiHUN:
We need to cache the number of players somehow, because this method will have the same result as `ESX.GetExtendedPlayers()`

Kenshiin13:
> We need to cache the number of players somehow, because this method will have the same result as `ESX.GetExtendedPlayers()`

Same result in what sense?

[core]/es_extended/server/functions.lua
Comment (thelindat) [reply [core]/es_extended/server/functions.lua:264]: > Odd. I would've expected them to be removed on disconnect.

They are? Hence the nil values.
It's a sparse array, so the length operator doesn't work.

thelindat:
> Also, you would have to trust developers to know that ESX.Players contains nil values and they can't just do `#ESX.GetExtendedPlayers()`.

... So the ESX team in their infinite wisdom _broke compatibility_? Who would have guessed.

mahanmoulaei:
> ... So the ESX team in their infinite wisdom _broke compatibility_? Who would have guessed.
> 
> [5e44373](https://github.com/esx-framework/esx_core/commit/5e44373d80c3eb94bad3b068788f5c742adb3e30)

Another one, not surprised



> The new ESX.GetNumPlayers function offers an efficient and straightforward approach to obtaining player counts based on specific attributes. By eliminating the need for a double loop,


Why introduce another function when all you have to do is to return table length as the second return value for the `ESX.GetExtendedPlayers()` 
__(ehem \*clears throat\* if the function wasn't broken) __

Gellipapa:
> 
I don't quite understand how you want to return the length of the table if you have a {police:{xPlayer,xPlayer,xPlayer,xPlayer,xPlayer}, sheriff:{xPlayer,xPlayer,xPlayer,xPlayer,xPlayer}}.
This is all time return 2, but i have 5 police, 5 sheriff xplayer in police, sheriff table.

At the moment there are several cases where you don't want the data of one job but the data of several jobs and why should you call several GetExtendedPlayers methods when you can solve it?

Kenshiin13:
> I don't quite understand how you want to return the length of the table if you have a {police:{xPlayer,xPlayer,xPlayer,xPlayer,xPlayer}, sheriff:{xPlayer,xPlayer,xPlayer,xPlayer,xPlayer}}. This is all time return 2, but i have 5 police, 5 sheriff xplayer in police, sheriff table.

You should be able to just get the length of each table. 

```lua
local xCops = ESX.GetExtendedPlayers("job", {"police", "sherrif"}
return (#(xCops.police) + #(xCops.sherrif))
```

Gellipapa:
> > I don't quite understand how you want to return the length of the table if you have a {police:{xPlayer,xPlayer,xPlayer,xPlayer,xPlayer}, sheriff:{xPlayer,xPlayer,xPlayer,xPlayer,xPlayer}}. This is all time return 2, but i have 5 police, 5 sheriff xplayer in police, sheriff table.
> 
> You should be able to just get the length of each table.
> 
> ```lua
> local xCops = ESX.GetExtendedPlayers("job", {"police", "sherrif"})
> return (#(xCops.police) + #(xCops.sherrif))
> ```

Yes, I believe this has been described before, it's a good solution, but I don't know how much the developers know. I see more than one place in paid scripts with a solution like what I wrote earlier, so it's not obvious at first, and they tend to double do a cycle. There should be an obvious solution where it is clear that yes there is a count.

Gellipapa:
It might be a good solution to return a second table in the second parameter and only count: {police:5,sheriff:5} and get the count and continue working with the xPlayer-elements. And the # count mentioned above would still work, but there would be an option to get count back immediately. 

For example: 
```lua
local xCops,count = ESX.GetExtendedPlayers("job", {"police", "sherrif"})
count.police -- 5
count.sherrif -- 5
```

How about such an intermediate solution?

Kenshiin13:
> but I don't know how much the developers know. I see more than one place in paid scripts with a solution like what I wrote earlier, so it's not obvious at first

I believe that will not fix this issue.

Gellipapa:
> I believe that will not fix this issue.

If they already have the option it's not our fault if they don't use it, you're right to expect this # counting, but that's how it is. That's why we need alternatives, because this little change I've written won't break the old systems and because it's just an extra variable if someone uses it and if someone doesn't we're there anyway. 

I think it can work.

Kenshiin13:
The only change I believe is remotely justified is a new function that simply returns a count. If you need to access the corresponding player objects anyways, you can just use `#` to get the count. There are a lot of instances where you may not care about the player objects and just want the count. For example:

- Listing player counts of important jobs in your HUD (EMS, PD)
- Disallowing the use of medikits/repairkits if enough EMS/LSC players are online.

Gellipapa:
> The only change I believe is remotely justified is a new function that simply returns a count. If you need to access the corresponding player objects anyways, you can just use `#` to get the count. There are a lot of instances where you may not care about the player objects and just want the count. For example:
> 
> * Listing player counts of important jobs in your HUD (EMS, PD)
> * Disallowing the use of medikits/repairkits if enough EMS/LSC players are online.

This is also true, but since you also have to count the number of ems etc. you go through the xPlayers in the background, from there it's just a matter of tabulating them and sending them back as one table and the count part in the second parameter. 

Because now if we look at your example and it needs the xPlayer and the count then my solution is better because it can solve it in one method and since you have to go through and count it anyway, I think it's no problem to put it in a table and send it through.

Kenshiin13:
We could just cache the job counts as previously suggested.

Gellipapa:
> We could just cache the job counts as previously suggested.

What do you mean? GlobalState?

Kenshiin13:
> What do you mean? GlobalState?

I'd prefer a table in ESX, that is only managed by ESX. But yeah, something like that.

Gellipapa:
> > What do you mean? GlobalState?
> 
> I'd prefer a table in ESX, that is only managed by ESX. But yeah, something like that.

Both are feasible and then you decide which one you want to use.

Kenshiin13:
> Both are feasible and then you decide which one you want to use.

I will push an udpate in the next couple of days, pretty swamped right now.

Gellipapa:
> > Both are feasible and then you decide which one you want to use.
> 
> I will push an udpate in the next couple of days, pretty swamped right now.

Thanks a lot if you do it, if you don't have time we will do it, just let us know.

Kenshiin13:
> Thanks a lot if you do it, if you don't have time we will do it, just let us know.

Just pushed a new commit, I believe that'd work

[core]/es_extended/server/functions.lua
```diff
@@ -261,24 +261,31 @@ end
...
+            end
+            return numPlayers
+        end
+    else
+        if key == "job" then
+            return (ESX.JobsPlayerCount[val] or 0)
+        else
+            local filteredPlayers = ESX.GetExtendedPlayers(key, val)
+            return #filteredPlayers
+        end
```
Comment (CsokiHUN) [inline [core]/es_extended/server/functions.lua:287]:
```suggestion
    if type(val) == "table" then
        local numPlayers = {}
        if key == "job" then
            for _, v in ipairs(val) do
                numPlayers[v] = (ESX.JobsPlayerCount[v] or 0)
            end
            return numPlayers
        end

        local filteredPlayers = ESX.GetExtendedPlayers(key, val)
        for i, v in pairs(filteredPlayers) do
            numPlayers[i] = (#v or 0)
        end
        return numPlayers
    end

    if key == "job" then
        return (ESX.JobsPlayerCount[val] or 0)
    end

    local filteredPlayers = ESX.GetExtendedPlayers(key, val)
    return #filteredPlayers

**Review by CsokiHUN — CHANGES_REQUESTED** [review]

File: [core]/es_extended/server/functions.lua (hunk 2 above)
Comment (Kenshiin13) [reply [core]/es_extended/server/functions.lua]: Thank you. I specifically used the returns to avoid nesting, but then forgot to remove the nesting.

Pushed the changes.

File: [core]/es_extended/server/functions.lua [hunk 3]
```diff
@@ -259,6 +259,36 @@ function ESX.GetExtendedPlayers(key, val)
...
+            numPlayers[i] = (#v or 0)
+        end
+        return numPlayers
+    end
+
+    if key == "job" then
+        return (ESX.JobsPlayerCount[val] or 0)
+    end
+
+    local filteredPlayers = ESX.GetExtendedPlayers(key, val)
```
Comment (Gellipapa) [inline [core]/es_extended/server/functions.lua:287]: #ESX.GetExtendedPlayers(key, val) because we don't use the filteredPlayers variable anywhere afterwards.

[core]/es_extended/server/functions.lua
```diff
@@ -259,6 +259,36 @@ function ESX.GetExtendedPlayers(key, val)
 	return xPlayers
 end
 
+function ESX.GetNumPlayers(key, val)
+    if(not key) then
```
Comment (Gellipapa) [inline [core]/es_extended/server/functions.lua:263]: remove please () only add  if not key then

[core]/es_extended/server/main.lua
```diff
@@ -377,6 +377,8 @@ AddEventHandler('playerDropped', function(reason)
 
 	if xPlayer then
 		TriggerEvent('esx:playerDropped', playerId, reason)
+        local job = xPlayer.getJob().name
+        ESX.JobsPlayerCount[job] = (ESX.JobsPlayerCount[job] or 1) -1
```
Comment (Gellipapa) [inline [core]/es_extended/server/main.lua:381]: Handle please minus values. 
For example:
ESX.JobsPlayerCount["police"] = 0
ESX.JobsPlayerCount["police"] = (ESX.JobsPlayerCount["police"] or 1) -1  

ESX.JobsPlayerCount["police"] -- returns 0 
ESX.JobsPlayerCount["police"] - 1 --  returns 0-1 = -1

[core]/es_extended/server/main.lua
```diff
@@ -385,6 +387,16 @@ AddEventHandler('playerDropped', function(reason)
 	end
 end)
 
+AddEventHandler("esx:playerLoaded", function(playerId, xPlayer, isNew)
+    local job = xPlayer.getJob().name
+    ESX.JobsPlayerCount[job] = (ESX.JobsPlayerCount[job] or 0) +1
+end)
+
+AddEventHandler("esx:setJob", function(src, job, lastJob)
+    ESX.JobsPlayerCount[lastJob.name] = (ESX.JobsPlayerCount[lastJob.name] or 1) - 1
```
Comment (Gellipapa) [inline [core]/es_extended/server/main.lua:396]: Handle please minus values.

Gellipapa (changes requested):
Please resolved request changes.

[core]/es_extended/server/functions.lua
Comment (Kenshiin13) [reply [core]/es_extended/server/functions.lua:287]: ok

[core]/es_extended/server/functions.lua
Comment (Kenshiin13) [reply [core]/es_extended/server/functions.lua:263]: ok

[core]/es_extended/server/main.lua
Comment (Kenshiin13) [reply [core]/es_extended/server/main.lua:381]: I'm not sure I understand what you mean.

[core]/es_extended/server/main.lua
Comment (Kenshiin13) [reply [core]/es_extended/server/main.lua:396]: I'm not sure I understand what you mean.

[core]/es_extended/server/main.lua
Comment (Kenshiin13) [reply [core]/es_extended/server/main.lua:381]: It tracks any users job, so how could someone still have a police job if the count is 0?

[core]/es_extended/server/main.lua
Comment (Kenshiin13) [reply [core]/es_extended/server/main.lua:396]: It tracks any users job, so how could someone still have a police job if the count is 0?

Kenshiin13:
Forgot to switch account for the last 2 commits, excuse me.

[core]/es_extended/server/main.lua
Comment (Kenshiin13) [reply [core]/es_extended/server/main.lua]: I have pushed a change anyways.

[core]/es_extended/server/main.lua
Comment (Kenshiin13) [reply [core]/es_extended/server/main.lua]: I have pushed a change anyways.

[core]/es_extended/server/main.lua
```diff
@@ -378,7 +378,7 @@ AddEventHandler('playerDropped', function(reason)
 	if xPlayer then
 		TriggerEvent('esx:playerDropped', playerId, reason)
         local job = xPlayer.getJob().name
-        ESX.JobsPlayerCount[job] = (ESX.JobsPlayerCount[job] or 1) -1
+        ESX.JobsPlayerCount[job] = ((ESX.JobsPlayerCount[job] and ESX.JobsPlayerCount[job] > 0) and ESX.JobsPlayerCount[job] or 1) -1
```
Comment (Gellipapa) [inline [core]/es_extended/server/main.lua:381]: It's okey, but please have a variable for it because you use it more than once.

For example
local currentJobCount = ESX.JobsPlayerCount[job]
ESX.JobsPlayerCount[job] = ((currentJobCount and currentJobCount > 0) and currentJobCount or 1) -1

[core]/es_extended/server/main.lua
```diff
@@ -393,7 +393,7 @@ AddEventHandler("esx:playerLoaded", function(playerId, xPlayer, isNew)
 end)
 
 AddEventHandler("esx:setJob", function(src, job, lastJob)
-    ESX.JobsPlayerCount[lastJob.name] = (ESX.JobsPlayerCount[lastJob.name] or 1) - 1
+    ESX.JobsPlayerCount[lastJob.name] = ((ESX.JobsPlayerCount[lastJob.name] and ESX.JobsPlayerCount[lastJob.name] > 0) and ESX.JobsPlayerCount[lastJob.name] or 1) -1
```
Comment (Gellipapa) [inline [core]/es_extended/server/main.lua:396]: It's okey, but please have a variable for it because you use it more than once.

For example
local currentJobCount = ESX.JobsPlayerCount[job]
ESX.JobsPlayerCount[job] = ((currentJobCount and currentJobCount > 0) and currentJobCount or 1) -1

Gellipapa (changes requested):
Please doing this little refactor and i can merge this pr.

Gellipapa:
#1202 It has been completed so this issue close.

PR #1170, opened 2023-08-21
Update commands.lua

Arctos2win:
this command would break if someone leaves and would  stop counting at the first nil value in the list. 

you can get the length of the table to get the online players message in front but that would be less performant.

### DIFF
--- a/[core]/es_extended/server/commands.lua
+++ b/[core]/es_extended/server/commands.lua
@@ -574,11 +574,11 @@ ESX.RegisterCommand("noclip", 'admin', function(xPlayer)
 end, false)
 
 ESX.RegisterCommand('players', "admin", function()
-	local xPlayers = ESX.GetExtendedPlayers() -- Returns all xPlayers
-	print(('^5%s^2 online player(s)^0'):format(#xPlayers))
-	for i = 1, #(xPlayers) do
-		local xPlayer = xPlayers[i]
-		print(('^1[^2ID: ^5%s^0 | ^2Name : ^5%s^0 | ^2Group : ^5%s^0 | ^2Identifier : ^5%s^1]^0\n'):format(
-			xPlayer.source, xPlayer.getName(), xPlayer.getGroup(), xPlayer.identifier))
-	end
+    local xPlayers = ESX.GetExtendedPlayers() -- Returns all xPlayers
+    local count = 0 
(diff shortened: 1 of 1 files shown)


Kenshiin13:
Haha, good catch!!

[core]/es_extended/server/commands.lua
```diff
@@ -574,11 +574,11 @@ ESX.RegisterCommand("noclip", 'admin', function(xPlayer)
...
-	local xPlayers = ESX.GetExtendedPlayers() -- Returns all xPlayers
-	print(('^5%s^2 online player(s)^0'):format(#xPlayers))
-	for i = 1, #(xPlayers) do
-		local xPlayer = xPlayers[i]
-		print(('^1[^2ID: ^5%s^0 | ^2Name : ^5%s^0 | ^2Group : ^5%s^0 | ^2Identifier : ^5%s^1]^0\n'):format(
-			xPlayer.source, xPlayer.getName(), xPlayer.getGroup(), xPlayer.identifier))
-	end
+    local xPlayers = ESX.GetExtendedPlayers() -- Returns all xPlayers
+    local count = 0 
+    for _, xPlayer in pairs(xPlayers) do
```
Comment (CsokiHUN) [inline [core]/es_extended/server/commands.lua:579]: You are wrong, the counting would not be interrupted because pairs are used instead of ipairs.
I do not consider this PR relevant

Comment (Kenshiin13) [reply [core]/es_extended/server/commands.lua:579]: > You are wrong, the counting would not be interrupted because `pairs` are used instead of `ipairs`. I do not consider this PR relevant

What do you mean? 
```lua
print(('^5%s^2 online player(s)^0'):format(#xPlayers))
```
Would result in wrong output as # is not nil value safe.

Comment (Arctos2win) [reply [core]/es_extended/server/commands.lua:579]: > 

i was talking about the code before, #xPlayers would break when someone leaves.

Arctos2win:
> Haha, good catch!!

Yeah got some complaints about this command in my pm xD

Kenshiin13:
> Yeah got some complaints about this command in my pm xD

Hopefully you also agree now that this should be mentioned in the docs xD

[core]/es_extended/server/commands.lua
Comment (CsokiHUN) [reply [core]/es_extended/server/commands.lua:579]: sorry, I confused the old version with the new one

CsokiHUN (approved):

PR #1173, opened 2023-08-21
Fixed the spawn of vehicle with "Utility" class

KadDarem:
The gas tank trailer replaces all vehicles with the "Utility" class (number 11). 
I fixed the spawn by removing the number 11 from the list to spawn these vehicles as "automobile" and not as "trailer".

### DIFF
--- a/[core]/es_extended/client/functions.lua
+++ b/[core]/es_extended/client/functions.lua
@@ -1355,7 +1355,7 @@ function ESX.GetVehicleType(model)
     local vehicleType = GetVehicleClassFromName(model)
     local types = {
         [8] = "bike",
-        [11] = "trailer",
+        --[11] = "trailer",
         [13] = "bike",
         [14] = "boat",
         [15] = "heli",
--- a/[core]/es_extended/client/functions.lua
+++ b/[core]/es_extended/client/functions.lua
@@ -1355,7 +1355,6 @@ function ESX.GetVehicleType(model)
     local vehicleType = GetVehicleClassFromName(model)
     local types = {
         [8] = "bike",
-        --[11] = "trailer",
         [13] = "bike",
         [14] = "boat",
         [15] = "heli",


NikartikTheBear:
From what i remember there is a pr from @Arctos2win  in the dev branch that solves the problem

KadDarem:
Yes. It's pretty similar

NikartikTheBear:
As an advice for future contributing, check the dev branch first because a problem may be already solved there and pull request to the dev branch not the main

MoskalykA:
My [pull request](https://github.com/esx-framework/esx_core/pull/1175) also seems to correct this (but I'm not sure).

KadDarem:
> My [pull request](https://github.com/esx-framework/esx_core/pull/1175) also seems to correct this (but I'm not sure).

It fixes it too. But it's boring to add one by one all vehicles with your way.

Gellipapa:
Hi Guys! if I understand correctly the problem here is that if the vehicleType is 11 then the tanker will spawn? 

Because there is a bug currently in car command that sometimes you enter a specific car so /car zentorno will sometimes spawn an orange tanker. 

Can you guys look into this and test it in this pr? 

Because this being commented out is not a solution, either delete or rework this part. 

Thanks.

KadDarem:
> Hi Guys! if I understand correctly the problem here is that if the vehicleType is 11 then the tanker will spawn? 
> 
> Because there is a bug currently in car command that sometimes you enter a specific car so /car zentorno will sometimes spawn an orange tanker. 
> 
> Can you guys look into this and test it in this pr? 
> 
> Because this being commented out is not a solution, either delete or rework this part. 
> 
> Thanks.

GetVehicleClassFromName return 11 for all vehicles with Utility class. So currently, they are all replaced by a gas tank trailer due to for ESX, 11 is trailer and not automobile.

You can't really rework this part due to trailer and utility return the same number. So you have two ways to fix it :
- Enter one by one the list of utility vehicles to overwrite the vehicle class.
- Spawn all vehicle with class 11 as automobile instead trailer.

I choose the second way to be compatible with add-on vehicles.

Gellipapa:
> > Hi Guys! if I understand correctly the problem here is that if the vehicleType is 11 then the tanker will spawn?
> > Because there is a bug currently in car command that sometimes you enter a specific car so /car zentorno will sometimes spawn an orange tanker.
> > Can you guys look into this and test it in this pr?
> > Because this being commented out is not a solution, either delete or rework this part.
> > Thanks.
> 
> GetVehicleClassFromName return 11 for all vehicles with Utility class. So currently, they are all replaced by a gas tank trailer due to for ESX, 11 is trailer and not automobile.
> 
> You can't really rework this part due to trailer and utility return the same number. So you have two ways to fix it :
> 
> * Enter one by one the list of utility vehicles to overwrite the vehicle class.
> * Spawn all vehicle with class 11 as automobile instead trailer.
> 
> I choose the second way to be compatible with add-on vehicles.

Hi! Thank you for the detailed description.

Then let's make it so that you delete it, not just comment on it, and I'll test how compatible it is with the old version.

Gellipapa (changes requested):

Gellipapa:
I have tested it to see if it works, please read my comment and fix it after I can merge.

KadDarem:
> @KadDarem I have tested it to see if it works, please read my comment and fix it after I can merge.

It's good. I removed the line

Gellipapa (approved):
Hi! Well done, I was tested and merged.

PR #1175, opened 2023-08-21
refactor: vehicle type without networking

MoskalykA:
Hello, the `GetVehicleClassFromName` function gives a reference list so I suggest not doing any more networking and just finding the right value on the server side.


Arctos2win:
> Hello, the `GetVehicleClassFromName` function gives a reference list so I suggest not doing any more networking and just finding the right value on the server side. 

what about addon vehicles

[core]/es_extended/server/functions.lua
```diff
@@ -287,21 +287,471 @@ function ESX.GetIdentifier(playerId)
...
-
-	if Core.vehicleTypesByModel[model] then
-		return cb(Core.vehicleTypesByModel[model])
-	end
-
-	ESX.TriggerClientCallback(player, "esx:GetVehicleType", function(vehicleType)
-		Core.vehicleTypesByModel[model] = vehicleType
-		cb(vehicleType)
-	end, model)
+local classFromName = {
```
Comment (Rav3n95) [inline [core]/es_extended/server/functions.lua:290]: Are you mental?

Comment (iSentrie) [reply [core]/es_extended/server/functions.lua:290]: jeez

MoskalykA:
> what about addon vehicles

I hadn't thought of that, but I could turn this pull request into a draft and make a pull request on fivem to implement server-side detection.
However, I find it unacceptable to be insulted on the pretext of a small error, especially coming from the leader of your team.

TheFantomas:
@MoskalykA Can close that? 

It is fixed in the latest version. It is not the best solution, but unfortunately there is no other solution for now, I would say.

PR #1178, opened 2023-08-22
Update server.cfg

WirzRexTon:
Why this, Whereas in es_extended, it says to delete it completely.


Gellipapa:
#1180 We fix issue this pr, but i merged because need remove these lines. Thank you.

PR #1181, opened 2023-08-23
Update commands.lua

t1ger-scripts:
Fixed command /players returning nil on xPlayer due to not correctly looped.


Cyzxin:
This command was already updated recently in [Dev Branch](https://github.com/esx-framework/esx_core/blob/dev/%5Bcore%5D/es_extended/server/commands.lua#L576-L584). If you still have proposed changes I would suggest making a new PR using dev branch

t1ger-scripts:
> This command was already updated recently in [Dev Branch](https://github.com/esx-framework/esx_core/blob/dev/%5Bcore%5D/es_extended/server/commands.lua#L576-L584). If you still have proposed changes I would suggest making a new PR using dev branch

Got it, no additional changes from here then. Thanks for quick reply.

PR #1185, opened 2023-08-24
Added check for spawned vehicles not on all wheels then it rotates

Manta95:


[core]/es_extended/client/functions.lua
```diff
@@ -430,7 +430,11 @@ function ESX.Game.SpawnVehicle(vehicleModel, coords, heading, cb, networked)
         while not HasCollisionLoadedAroundEntity(vehicle) do
             Wait(0)
         end
-
+        if not IsVehicleOnAllWheels(vehicle) then
+		    local carCoords = GetEntityRotation(vehicle, 2)
+		    Wait(500)
+		    SetEntityRotation(vehicle, carCoords[1], 0, carCoords[3], 2, true)
```
Comment (Gellipapa) [inline [core]/es_extended/client/functions.lua:436]: Is this 1-3 index all time?

Comment (Manta95) [reply [core]/es_extended/client/functions.lua:436]: Yep, tried with all kind of vehicles (mostly addons trailers, boats and planes) and it works well

You can try with this trailer who spawn upside-down https://www.gta5-mods.com/vehicles/brute-camp-trailer-addon

Manta95:
> @Manta95 Hi! This is not working, I can't accept this because it requires a newer artifact or gamebuild version. This should include some condition testing to make it work if all conditions are met.
> 
> (Only test /car command and get this in server console) image

Ok, it's not about artifact nor gambuild, it's the native that didn't work server side. I try to find a fix for this

Gellipapa:
> > @Manta95 Hi! This is not working, I can't accept this because it requires a newer artifact or gamebuild version. This should include some condition testing to make it work if all conditions are met.
> > (Only test /car command and get this in server console) image
> 
> Ok, it's not about artifact nor gamebuild, it's the native that didn't work server side. I try to find a fix for this

I was thinking that under a given version it is not native yet, that's why I wrote this, but it could be a different problem.

PR #1195, opened 2023-09-01
Improve the way the identifier is fetched

Arctos2win:


Gellipapa:
@Arctos2win Hi! I don't know how reliable it is native, I can't really find any description of it, do you have any description of it?

Kenshiin13:
> @Arctos2win Hi! I don't know how reliable it is native, I can't really find any description of it, do you have any description of it?

Here is the [native documentation](https://docs.fivem.net/natives/?_0xA61C8FC6).

Gellipapa:
> > @Arctos2win Hi! I don't know how reliable it is native, I can't really find any description of it, do you have any description of it?
> 
> Here is the [native documentation](https://docs.fivem.net/natives/?_0xA61C8FC6).

Hi! Thanks but i found this topic, but no one has used it before so I would need a more definitive description for a very long time with the GetIdentifiers for loop was solution.

Does it support all license types? discord,xbl, etc?

Kenshiin13:
> Hi! Thanks but i found this topic, but no one has used it before so I would need a more definitive description for a very long time with the GetIdentifiers for loop was solution.
> 
> Does it support all license types? discord,xbl, etc?

Honestly, I think under the hood it does the same thing as looping over the identifiers and matching a string. See the return value:

> Returns:
The identifier that matches the string provided

[core]/esx_multicharacter/server/main.lua
```diff
@@ -31,17 +31,13 @@ local PREFIX = Config.Prefix or 'char'
...
-			identifier = string.gsub(v, identifier, '')
-			return identifier
-		end
-	end
+    local fxDk = GetConvarInt('sv_fxdkMode', 0)
+    if fxDk == 1 then
+        return "ESX-DEBUG-LICENCE"
+    end
+    
+    local identifier = string.gsub(GetPlayerIdentifierByType(source, PRIMARY_IDENTIFIER), PRIMARY_IDENTIFIER .. ':', '')
```
Comment (Gellipapa) [inline [core]/esx_multicharacter/server/main.lua:39]: Okey, i tested it's works, but i don't have steam identifier and return nil but string.gsub don't handle nil value we needs to handle nil values.

Gellipapa (changes requested):

Gellipapa:
> > Hi! Thanks but i found this topic, but no one has used it before so I would need a more definitive description for a very long time with the GetIdentifiers for loop was solution.
> > Does it support all license types? discord,xbl, etc?
> 
> Honestly, I think under the hood it does the same thing as looping over the identifiers and matching a string. See the return value:
> 
> > Returns:
> > The identifier that matches the string provided

Yes i tested, now and working. Thank you for help.

[core]/esx_multicharacter/server/main.lua
```diff
@@ -36,8 +36,14 @@ local function GetIdentifier(source)
...
-    local identifier = string.gsub(GetPlayerIdentifierByType(source, PRIMARY_IDENTIFIER), PRIMARY_IDENTIFIER .. ':', '')  
-    return identifier 
+    local identifier = GetPlayerIdentifierByType(source, PRIMARY_IDENTIFIER)
+
+    if not identifier then
+        return
+    end 
+    
+    identifier = identifier:gsub(PRIMARY_IDENTIFIER .. ':', '')
+    return identifier
```
Comment (Gellipapa) [inline [core]/esx_multicharacter/server/main.lua:46]: return identifier:gsub(PRIMARY_IDENTIFIER .. ':', '') enough, don't need pass to variable, you don't use identifier afterwards.

Gellipapa (approved):
Please fix little refactor and i can merge. Thank you.

[core]/esx_multicharacter/server/main.lua
Comment (Arctos2win) [reply [core]/esx_multicharacter/server/main.lua:46]: i did that to avoid the extra number behind the identifier that would be returned

Comment (Gellipapa) [reply [core]/esx_multicharacter/server/main.lua:46]: > i did that to avoid the extra number behind the identifier that would be returned

It's okey, i forget gsub return 2 values.

PR #1197, opened 2023-09-02
move registernetevent

Arctos2win:
this commit moves the registernetevent 
some people that are using esx_multicharacter might have the problem that its unsecure for the net.

### DIFF
--- a/[core]/es_extended/server/main.lua
+++ b/[core]/es_extended/server/main.lua
@@ -19,6 +19,7 @@ end
 
 loadPlayer = loadPlayer .. ' FROM `users` WHERE identifier = ?'
 
+RegisterNetEvent('esx:onPlayerJoined') 
 if Config.Multichar then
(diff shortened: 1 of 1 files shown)


Gellipapa:
@Arctos2win Hi! I won't accept this because it's a server-side event, don't call it in the client. It contains data that is server-side. That's probably why it's not registered on the network so that only the server side can access it and it's safe.

When you get a "not safe for net" error when calling it, it's because it was called on the wrong page. You need to register an event if you want to trigger an event between server->client or client->server. 

In this case it's completely server side so you don't need to register the event on the network because the server can use this event on the other server sides.

Gellipapa:
@Arctos2win Hi! What do you think about the above?

Arctos2win:
> @Arctos2win Hi! What do you think about the above?

Hey yes i looked at it better and idd its done a weird way but i know why its considered insecure
to register it when using multicharacter

PR #1199, opened 2023-09-03
improve fetching identifier in es_extended ( same way as in multichar )

Arctos2win:


Gellipapa:
@Arctos2win Hi! Well done, I tested.

PR #1212, opened 2023-09-17
fix stop do not use resources

Arctos2win:


[core]/es_extended/server/main.lua
```diff
@@ -748,16 +748,20 @@ local DoNotUse = {
     ['default_spawnpoint'] = true,
 }
 
+AddEventHandler('onResourceStart', function(key)
+    if DoNotUse[string.lower(key)] then
+        while GetResourceState(key) ~= 'started' do 
+            Wait()
```
Comment (Gellipapa) [inline [core]/es_extended/server/main.lua:754]: Please add 0 i believe works if default value is 0, but we all code use parameter. Please change to 0.

Gellipapa (changes requested):

Gellipapa (approved):

PR #1217, opened 2023-09-17
refactor(Player Class): Typing Code & Perf

Walker974:
268528063-d3191c7e-b258-4423-b73b-1204f47c2d1c


Thekuca (commented):
I don't see the reason for parentheses IMO the code is cleaner without them. Other then that it's all good.

Walker974:
Hey, yeah i agree with you.
Thank you
Have a good day

benzon:
Wondering why you are using
local _GetPlayerPed = GetPlayerPed;
Instead of just local GetPlayerPed = GetPlayerPed

And why ; is added in the end, looks messy.

Walker974:
Yes you can just use  local GetPlayerPed = GetPlayerPed
For ; in the end, it is just because I am familiar with C

Walker974:
Hello @Gellipapa ! 
Good !

Kenshiin13:
> Hi! @Walker974 Please remove ; and () in code. Keep the lua simple.

I can get behind not using semicolons, but I must say I find code less readable without '()'. Obviously, that is subjective.

benzon:
> > Hi! @Walker974 Please remove ; and () in code. Keep the lua simple.
> 
> I can get behind not using semicolons, but I must say I find code less readable without '()'. Obviously, that is subjective.

Think it comes down to coding preferences and styles in the end, but it needs to be consistent throughout a project, and not end up having different coding styles, and formatting.

benzon:
> Yes you can just use local GetPlayerPed = GetPlayerPed For ; in the end, it is just because I am familiar with C

Personally, I would prefer that natives are not redeclared (loaded) and then set to a new variable.

local GetPlayerPed = GetPlayerPed over local _GetPlayerPed = GetPlayerPed

Kenshiin13:
>> 
> Think it comes down to coding preferences and styles in the end, but it needs to be consistent throughout a project, and not end up having different coding styles, and formatting.

Of course, I am not suggesting a change, everyone should stick to the style guidelines of the project.

Thekuca (approved):
Looks good

PR #1232, opened 2023-09-29
improve multiple functions and one event

Arctos2win:


[core]/es_extended/client/functions.lua
```diff
@@ -420,20 +402,19 @@ function ESX.Game.GetObjects() -- Leave the function for compatibility
...
-
-    if not onlyOtherPeds then
-        return pool
-    end
-
-    local peds = {}
-    for i = 1, #pool do
-        if pool[i] ~= myPed then
-            peds[#peds + 1] = pool[i]
+    local pool = GetGamePool('CPed')
```
Comment (Gellipapa) [inline [core]/es_extended/client/functions.lua:405]: I don't think it's lucky that the full pool is returned by default. Because then compatibility is broken because the default is to return an empty table, which is what the old method did. Now it will always return the pool. So it won't be backwards compatible.

[core]/es_extended/client/functions.lua
```diff
@@ -442,18 +423,19 @@ end
 
 function ESX.Game.GetPlayers(onlyOtherPlayers, returnKeyValue, returnPeds)
     local players, myPlayer = {}, PlayerId()
+    local active = GetActivePlayers() 
 
-    for _, player in ipairs(GetActivePlayers()) do
-        local ped = GetPlayerPed(player)
+    for i=1, #active do 
+        local ped = GetPlayerPed(active[i])
```
Comment (Gellipapa) [inline [core]/es_extended/client/functions.lua:429]: I think you should pass any variable active[i]. For example:  local currentPlayer = active[i] and use currentPlayer all usages.

Gellipapa (changes requested):

PR #1246, opened 2023-10-15
feat(ESX.Players): metamethod for counting players

Thekuca:
Added a metamethod to the ESX.Players table to count players nil proof with #ESX.Players


Gellipapa:
@Thekuca Hi! Personally, I don't think it's a good idea to introduce this because it might confuse people, it won't be consistent. Users know how ESX.Players works and what helper methods can be used to extract information from it. Later on this can be an unforeseen problem if you override the basic len property. 

So I wouldn't do this in Legacy, but such solutions in Infinity which is a completely new system can be integrated there in the future.

Thekuca:
> @Thekuca Hi! Personally, I don't think it's a good idea to introduce this because it might confuse people, it won't be consistent. Users know how ESX.Players works and what helper methods can be used to extract information from it. Later on this can be an unforeseen problem if you override the basic len property.
> 
> So I wouldn't do this in Legacy, but such solutions in Infinity which is a completely new system can be integrated there in the future.

I agree with you.

PR #1258, opened 2023-11-28
Create id.lua

IAmNotGod-core:
Add Indonesian translations


Arctos2win:
@IAmNotGod-core what language is this ?

PR #1259, opened 2023-11-29
Refactoring code for Scaleform movie queries and 3D text drawing

Riyane3326:
This commit makes significant changes to improve the structure of the code. The Scaleform movie query function has been moved directly into the 'scaleform.lua' module, eliminating the need to access it via an unnecessary 'Utils' object. Similarly, the 3D text drawing function has been moved to the 'ESX.Game' object, improving the accessibility and simplicity of the code structure.

These changes are aimed at streamlining access to specific functionality, reducing complexity and making the code clearer and easier to maintain. The adjustments made ensure a better organization of functionalities, facilitating future development and maintenance.

### DIFF
--- a/[core]/es_extended/client/functions.lua
+++ b/[core]/es_extended/client/functions.lua
@@ -9,11 +9,7 @@ ESX.UI.Menu.RegisteredTypes = {}
 ESX.UI.Menu.Opened = {}
 
 ESX.Game = {}
-ESX.Game.Utils = {}
-
 ESX.Scaleform = {}
-ESX.Scaleform.Utils = {}
-
 ESX.Streaming = {}
 
 function ESX.IsPlayerLoaded()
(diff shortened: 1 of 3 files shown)


Arctos2win:
This pr would break backwards compatibility so its not very likely to be merged. Maybe @Gellipapa thinks different abt this but i dont think so.

Gellipapa:
@Riyane3326 Hi! Thank you for your PR and your work but unfortunately we can't accept this. The scaleform part is not so much risky because people don't call that method by itself, but the DrawText3D out of the Utils table breaks the code 100% and a fresh es_extended afterwards would give the user the error that there is no such method as Utils.DrawText3D so we can't accept that now.

It is important to keep backwards compatibility as much as we can and so unfortunately such core methods cannot be refactored.

PR #1264, opened 2023-12-08
refactor: add an `esx:onCommand` event

MoskalykA:
Hi
I suggest adding an `esx:onCommand` event because the event that retrieves messages from the chat doesn't do that.

### DIFF
--- a/[core]/es_extended/server/functions.lua
+++ b/[core]/es_extended/server/functions.lua
@@ -35,8 +35,9 @@ function ESX.RegisterCommand(name, group, cb, allowConsole, suggestion)
 	Core.RegisteredCommands[name] = { group = group, cb = cb, allowConsole = allowConsole, suggestion = suggestion }
 
 	RegisterCommand(name, function(playerId, args)
-		local command = Core.RegisteredCommands[name]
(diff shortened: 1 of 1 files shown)


MoskalykA:
> Hi! What and when would you use this event for? Personally I don't see the need for it.

Hi, for a logging system, to find out if a staff is abusing these permissions or if a player is pretending not to have done something that does

Gellipapa:
> > Hi! What and when would you use this event for? Personally I don't see the need for it.
> 
> Hi, for a logging system, to find out if a staff is abusing these permissions or if a player is pretending not to have done something that does

I see, but here if you don't have the right to do it then the code block won't run and the event won't be triggered, and if it doesn't run and the permission check is done later then the log is false because it says that I typed noclip, but I can't noclip it because it returns false at the end and won't call the noclip method.

MoskalykA:
> I see, but here if you don't have the right to do it then the code block won't run and the event won't be triggered, and if it doesn't run and the permission check is done later then the log is false because it says that I typed noclip, but I can't noclip it because it returns false at the end and won't call the noclip method.

I will modify my code

PR #1265, opened 2023-12-08
refactor: unused varargs

MoskalykA:
Hi, I suggest deleting this vararg which seems to be unused

### DIFF
--- a/[core]/es_extended/server/functions.lua
+++ b/[core]/es_extended/server/functions.lua
@@ -455,12 +455,12 @@ function ESX.RegisterUsableItem(item, cb)
 	Core.UsableItemsCallbacks[item] = cb
 end
 
-function ESX.UseItem(source, item, ...)
+function ESX.UseItem(source, item)
 	if ESX.Items[item] then
 		local itemCallback = Core.UsableItemsCallbacks[item]
 
 		if itemCallback then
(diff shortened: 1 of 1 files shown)


Gellipapa:
@MoskalykA Hi! What happens if somebody wants to pass some property of the item, or some extra variable that they want to get back in the callback?

```lua
function useHealthKit(source, item, healthRestoreAmount)
    print("Health kit used by player " .. source .. ". Restores " .. healthRestoreAmount .. " health.")
end

Core.UsableItemsCallbacks["healthkit"] = useHealthKit
```

```lua
ESX.UseItem(123, "healthkit", 50)
```

Gellipapa:
So clearly we cannot accept it because we use it.

MoskalykA:
I agree, although I haven't found any use for it ([search](https://github.com/search?q=fivem+ESX.UseItem&type=code)), otherwise you'd have to use a table rather than a vargarg, and there should be a distinct difference in performance.

Gellipapa:
> I agree, although I haven't found any use for it ([search](https://github.com/search?q=fivem+ESX.UseItem&type=code)), otherwise you'd have to use a table rather than a vargarg, and there should be a distinct difference in performance.

Yes in the past a table might have been better but now you can still pass a table because of this operator, but if you don't add a table and overwrite or delete it you break the system.

PR #1296, opened 2024-01-27
Fix VehicleProperties function and remove credits

Arctos2win:


Gellipapa:
@Arctos2win Have you tested this? Because it's funny how cumbersome it used to be, I know you said it didn't work, but does it work now?

Arctos2win:
didn't say it didn't work it doesn't work as its intended, 
statebags apply everytime you come in the scope of an entity that has statebags. 
so checking if the entity exists is completly unnecessary since the handler wouldn't be triggered if it didn't exist

PR #1298, opened 2024-01-27
Remove or statements that don't do stuff

Arctos2win:


Gellipapa (approved):
It's okey, if you don't add notifyType or length default nil and override param yes this is not working. Thanks for removed old code.

PR #1299, opened 2024-01-27
minimize DoesJobExist function

Arctos2win:


Gellipapa (approved):
Good job, it's works, and simple solution. Thank you.

PR #1303, opened 2024-01-29
refactor(es_extended/client/functions.lua) and (es_extended/client/modules/streaming.lua)

feelfreetofee:
Big refactor to client functions with no breaking changes (allegedly).
Except for ESX.Game.SpawnObject that returns the object synchronous if no callback provided.
(Just a new behaviour, keeps working as used to)

Added new functions:
 · ESX.DrawMissionText
 · ESX.Game.GetShapeTestResultSync **
 · ESX.Game.RaycastScreen **
** Last two are used to improve the performance of ESX.Game.GetVehicleInDirection
It already uses a Synchronous Shape Test Result less performant than solution provided.

In streaming functions, ff no callback provided, functions will return the first argument synchronously.
if not, first argument will be passed to the callback.

### DIFF
--- a/[core]/es_extended/client/functions.lua
+++ b/[core]/es_extended/client/functions.lua
@@ -24,19 +24,31 @@ function ESX.GetPlayerData()
     return ESX.PlayerData
 end
 
+local function IsResourceFound(resource)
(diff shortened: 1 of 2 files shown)


bitpredator (approved):

Gellipapa (approved):
Hi! @feelfreetofee and @Thekuca  Thanks for this client side refactor, i merged.

PR #1309, opened 2024-02-09
feat: add TypeCheck to check value types

CENSOR1337:
Add a utility function that allow to check value types

as I say in source code
Need feedback on this one, should we assert or return false and error message?
I prefer assert, but I'm open to suggestions

Thank you.

Here an examples

Example: 1
```lua
local value = 1
ESX.TypeCheck(value, "number") -- true
```

Example: 2
```lua
local value = "hello"
ESX.TypeCheck(value, "number", "string") -- true
```

Example: 3
```lua
local value = "hello"
ESX.TypeCheck(value, "number") -- false, "bad value (number expected, got string)"
```

Example: 4
```lua
local value = "hello"
ESX.TypeCheck(value, "boolean", "number", "table") -- false, "bad value (boolean or number or table expected, got string)"
```

Example: 5
```lua
local value = nil
ESX.TypeCheck(value, "boolean", "number", "table") -- false, "bad value (boolean or number or table expected, got nil)"
```

Example: 6
```lua
local value = nil
ESX.TypeCheck(value, "boolean", "number", "table", "nil") -- true
```

### DIFF
--- a/[core]/es_extended/common/functions.lua
+++ b/[core]/es_extended/common/functions.lua
@@ -106,3 +106,25 @@ end
 function ESX.Round(value, numDecimalPlaces)
     return ESX.Math.Round(value, numDecimalPlaces)
 end
+
+function ESX.TypeCheck(value, ...)
+    local types = { ... }
+
+    if (#types == 0) then return true end
+
+    local mapType = {}
+    for i = 1, #types, 1 do
+        mapType[types[i]] = true
+    end
+
+    local valueType = type(value)
+    local requireTypes = table.concat(types, " or ")
+    local errorMessage = ("bad value (%s expected, got %s)"):format(requireTypes, valueType)
+    local matches = mapType[valueType] ~= nil
+
+    -- Need feedback on this one, should we assert or return false and error message?
+    -- i prefer assert, but i'm open to suggestions
+    assert(matches, errorMessage)
+
+    return matches, errorMessage
+end
--- a/[core]/es_extended/common/functions.lua
+++ b/[core]/es_extended/common/functions.lua
@@ -107,24 +107,35 @@ function ESX.Round(value, numDecimalPlaces)
     return ESX.Math.Round(value, numDecimalPlaces)
 end
 
-function ESX.TypeCheck(value, ...)
+function ESX.ValidateType(value, ...)
     local types = { ... }
-
     if (#types == 0) then return true end
 
     local mapType = {}
     for i = 1, #types, 1 do
-        mapType[types[i]] = true
+        local validateType = types[i]
+        assert(type(validateType) == "string", "bad argument types, only expected string") -- should never use anyhing else than string
+        mapType[validateType] = true
     end
 
     local valueType = type(value)
-    local requireTypes = table.concat(types, " or ")
-    local errorMessage = ("bad value (%s expected, got %s)"):format(requireTypes, valueType)
-    local matches = mapType[valueType] ~= nil
 
-    -- Need feedback on this one, should we assert or return false and error message?
-    -- i prefer assert, but i'm open to suggestions
+    local matches = (mapType[valueType] ~= nil)
+
+    if not (matches) then
+        local requireTypes = table.concat(types, " or ")
+        local errorMessage = ("bad value (%s expected, got %s)"):format(requireTypes, valueType)
+
+        return false, errorMessage
+    end
+
+    return true
+end
+
+function ESX.AssertType(...)
+    local matches, errorMessage = ESX.ValidateType(...)
+
     assert(matches, errorMessage)
 
-    return matches, errorMessage
+    return matches
 end
--- a/[core]/es_extended/common/functions.lua
+++ b/[core]/es_extended/common/functions.lua
@@ -109,7 +109,7 @@ end
 
 function ESX.ValidateType(value, ...)
     local types = { ... }
-    if (#types == 0) then return true end
+    if #types == 0 then return true end
 
     local mapType = {}
     for i = 1, #types, 1 do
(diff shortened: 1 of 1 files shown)


CENSOR1337:
With this function we will able to improve readability of coude
Example setMeta function
from 
```lua
function self.setMeta(index, value, subValue)
    if not index then
        return print("[^1ERROR^7] xPlayer.setMeta ^5index^7 is Missing!")
    end

    if type(index) ~= "string" then
        return print("[^1ERROR^7] xPlayer.setMeta ^5index^7 should be ^5string^7!")
    end

    if value == nil then
        return print("[^1ERROR^7] xPlayer.setMeta value is missing!")
    end

    local _type = type(value)

    if not subValue then
        if _type ~= "number" and _type ~= "string" and _type ~= "table" then
            return print(("[^1ERROR^7] xPlayer.setMeta ^5%s^7 should be ^5number^7 or ^5string^7 or ^5table^7!"):format(value))
        end

        self.metadata[index] = value
    else
        if _type ~= "string" then
        end

        if not self.metadata[index] or type(self.metadata[index]) ~= "table" then
            self.metadata[index] = {}
        end

        self.metadata[index] = type(self.metadata[index]) == "table" and self.metadata[index] or {}
        self.metadata[index][value] = subValue
    end

    Player(self.source).state:set("metadata", self.metadata, true)
end
```
to 
```lua
function self.setMeta(index, value, subValue)
    ESX.TypeCheck(index, "string")

    if not subValue then
        ESX.TypeCheck(value, "booleam", "number", "string", "table")

        self.metadata[index] = value
    else
        ESX.TypeCheck(value, "string")

        ESX.TypeCheck(subValue, "booleam", "number", "string", "table")

        if not self.metadata[index] or type(self.metadata[index]) ~= "table" then
            self.metadata[index] = {}
        end

        self.metadata[index] = type(self.metadata[index]) == "table" and self.metadata[index] or {}
        self.metadata[index][value] = subValue
    end

    Player(self.source).state:set("metadata", self.metadata, true)
end
```

CENSOR1337:
Due to last commit, I decided to removed `TypeCheck` and add `AssertType` and `ValidateType` instead

```
 - replaced `TypeCheck` with `AssertType` this function will assert
 - add : `ValidateType` this function will return Boolean and error message
```

Here is an update code
```lua
function self.setMeta(index, value, subValue)
    ESX.AssertType(index, "string")

    if not subValue then
        ESX.AssertType(value, "number", "string", "table")

        self.metadata[index] = value
    else
        ESX.AssertType(value, "string")

        ESX.AssertType(subValue, "number", "string", "table")

        if not self.metadata[index] or type(self.metadata[index]) ~= "table" then
            self.metadata[index] = {}
        end

        self.metadata[index] = type(self.metadata[index]) == "table" and self.metadata[index] or {}
        self.metadata[index][value] = subValue
    end

    Player(self.source).state:set("metadata", self.metadata, true)
end
```

Example of `ValidateType`
```lua
local value = 1

if (ESX.ValidateType(value, "number")) then
    print("value is a number")
else
    print("value is not a number")
end

```

Gellipapa:
Hi! @CENSOR1337  I like this kind of solution, but it would be nice if we could also support ENUM types. I'm thinking here I'll give a table string and if it matches one of those then it's the right type but thanks for pr.

How do you check the boolean types, if I type 1 is it now number or boolean? 

It would be nice to have a way to determine from a string "true" or 1 whether it is a boolean value.

The latter might not be used, it's just an idea.

What do you think about this?

CENSOR1337:
> Hi! @CENSOR1337 I like this kind of solution, but it would be nice if we could also support ENUM types. I'm thinking here I'll give a table string and if it matches one of those then it's the right type but thanks for pr.
> 
> How do you check the boolean types, if I type 1 is it now number or boolean?
> 
> It would be nice to have a way to determine from a string "true" or 1 whether it is a boolean value.
> 
> The latter might not be used, it's just an idea.
> 
> What do you think about this?

Hey, that's great to hear! 
I think if we want to add support for other data types with specific conditions, it might be better to create separate functions for. This keeps the current function focused on providing type safety for the core Lua data types: boolean, function, nil, number, string, and table.

Regarding your question about checking for boolean types specifically:
`How do you check the boolean types, if I type 1 is it now number or boolean?`
```lua
local a_simple_value = 1
if (ESX.ValidateType(a_simple_value, "number", "boolean")) then
    print("a_simple_value is a number or a boolean")
else
    print("a_simple_value is not a number or a boolean")
end
```

Gellipapa:
> > Hi! @CENSOR1337 I like this kind of solution, but it would be nice if we could also support ENUM types. I'm thinking here I'll give a table string and if it matches one of those then it's the right type but thanks for pr.
> > How do you check the boolean types, if I type 1 is it now number or boolean?
> > It would be nice to have a way to determine from a string "true" or 1 whether it is a boolean value.
> > The latter might not be used, it's just an idea.
> > What do you think about this?
> 
> Hey, that's great to hear! I think if we want to add support for other data types with specific conditions, it might be better to create separate functions for. This keeps the current function focused on providing type safety for the core Lua data types: boolean, function, nil, number, string, and table.
> 
> Regarding your question about checking for boolean types specifically: `How do you check the boolean types, if I type 1 is it now number or boolean?` using this function I believed using this example would be a great choice
> 
> ```lua
> local a_simple_value = 1
> if (ESX.ValidateType(a_simple_value, "number", "boolean")) then
>     print("a_simple_value is a number or a boolean")
> else
>     print("a_simple_value is not a number or a boolean")
> end
> ```

Hi! Yes,it's legitimate not to complicate it in the first place, it's also a big enough help to write more readable code. Thanks your pr.

Gellipapa (changes requested):
Please remove brackets in code, lua don't needed and i can merge. Thanks.

CENSOR1337:
> Please remove brackets in code, lua don't needed and i can merge. Thanks.

I have removed some of brackets, Let me know what you think.
Thanks.

Gellipapa (approved):
Thanks, i will test and merge.

Gellipapa:
@CENSOR1337 Hi! Thanks i merged.

CENSOR1337:
> @CENSOR1337 Hi! Thanks i merged.

Thank you.

PR #1311, opened 2024-02-12
Update actions.lua

Marlox4:
Add ESX.PlayerData.seat and ESX.PlayerData.vehicle


Gellipapa:
Hi! @Marlox4 

This can be useful, but avoid duplication of code. Could you write a wrapper method that you call in two places with two values and if parameter nil then false so you handle two states in one method.

Can you do this?

Gellipapa:
@Marlox4 Hi! I fixed and refactored your solution, thanks for this feature, merged.

PR #1315, opened 2024-02-14
Update main.lua

WhereiamL:
Added the permanent ID option.

Usage

print(LocalPlayer.state.permID)


Mycroft-Studios:
this is just super not needed tbh, literally 0 usecase for it, and having it purely for statebags (which the client can change) is pretty pointless

Gellipapa:
Hi! Thank you for the pull request, but we will not merge it as there is no case where it is useful, it is a unique case that is useful for 1 out of 100 servers but thanks for pr.

PR #1316, opened 2024-02-16
Create markers.lua

fusion-develops:
For ESX point system simpler version of ox libs


Gellipapa:
Hi! @fusion-develops 

I like your solution and thanks for pr, but I don't know how happy Linden would be to use exactly the same name, so he's really referring to the ox_lib points function. 

If we don't look at this part because we are, say, just drawing inspiration from ox_lib, then we have name convention problems and probably the code doesn't work because it is used somewhere "inside" or "Inside" something in code.

Have you tested this code?

fusion-develops:
> Hi! @fusion-develops
> 
> I like your solution and thanks for pr, but I don't know how happy Linden would be to use exactly the same name, so he's really referring to the ox_lib points function.
> 
> If we don't look at this part because we are, say, just drawing inspiration from ox_lib, then we have name convention problems and probably the code doesn't work because it is used somewhere "inside" or "Inside" something in code.
> 
> Have you tested this code?

I've redone it and fixed the bugs and changed the name to Markers since that's what it would mostly be used for once approved I can use this to redo all the add-on resources

[core]/es_extended/client/modules/points.lua
```diff
@@ -0,0 +1,60 @@
...
+end
+
+CreateThread(function()
+    while true do 
+        Wait(0)
+        local InPoint = false
+        local Coords = GetEntityCoords(ESX.PlayerData.Ped)
+        for i = 1, #Markers do
+            local Marker = Markers[i]
+            Marker.distance = #(Coords - Marker.Coords)
```
Comment (Gellipapa) [inline [core]/es_extended/client/modules/points.lua:17]: Marker.Coords doesn't exist, and you should handle vector3. I mean convert if not vector 3.

[core]/es_extended/client/modules/points.lua
```diff
@@ -0,0 +1,60 @@
...
+CreateThread(function()
+    while true do 
+        Wait(0)
+        local InPoint = false
+        local Coords = GetEntityCoords(ESX.PlayerData.Ped)
+        for i = 1, #Markers do
+            local Marker = Markers[i]
+            Marker.distance = #(Coords - Marker.Coords) 
+            if Marker.distance < Marker.radius then 
+                if Marker.status and Marker.Inside then
```
Comment (Gellipapa) [inline [core]/es_extended/client/modules/points.lua:19]: Marker.Inside, Marker.inside bad variable name, and don't use. You dont't set Marker.inside.

[core]/es_extended/client/modules/points.lua
```diff
@@ -0,0 +1,60 @@
...
+function ESX.CreateMarker(data) 
+    local Index = #Points+1
+    Markers[Index] = data
+    return Markers[Index]
+end
+
+CreateThread(function()
+    while true do 
+        Wait(0)
+        local InPoint = false
```
Comment (Gellipapa) [inline [core]/es_extended/client/modules/points.lua:13]: Change all variable is correct lowerCamelCase. Thanks

Gellipapa (changes requested):
Please test your code before open PR. Thanks a lot.

[core]/es_extended/client/modules/points.lua
Comment (fusion-develops) [reply [core]/es_extended/client/modules/points.lua]: Done

[core]/es_extended/client/modules/points.lua
Comment (fusion-develops) [reply [core]/es_extended/client/modules/points.lua]: Fixed Marker.Inside is now a check if the function exists

[core]/es_extended/client/modules/points.lua
Comment (fusion-develops) [reply [core]/es_extended/client/modules/points.lua]: Removed Variable for a better way to do what it was meant for

[core]/es_extended/client/modules/markers.lua
```diff
@@ -0,0 +1,101 @@
...
+    while true do 
+        Wait(1500)
+        local Coords = GetEntityCoords(ESX.PlayerData.Ped)
+        for i = 1, #Markers do
+            local Marker = Markers[i]
+            Marker.distance = #(Coords - Marker.coords) 
+            if Marker.distance < Marker.radius then 
+                if Marker.status and Marker.Inside and not Marker.running then 
+                    Marker.running = true
+                    Markers:Run(i)
```
Comment (Gellipapa) [inline [core]/es_extended/client/modules/markers.lua:67]: hm, if i have 5 markers i created 5 thread?  I think we should avoid this and run code once thread.

Gellipapa (changes requested):

[core]/es_extended/client/modules/markers.lua
Comment (Gellipapa) [reply [core]/es_extended/client/modules/markers.lua:67]: and please change all variable is correct lowerCamelCase. Thanks

Comment (fusion-develops) [reply [core]/es_extended/client/modules/markers.lua]: Fixed both brother

Gellipapa (commented):
Do you tested this refactored code? Probably the while loop will cause you to stop the rest of the code from running if there is any.

Gellipapa:
Closed due to inactivity.

PR #1318, opened 2024-02-19
feat(identity): ui redesign

nwvh:
# UI Redesign PR
Quick, Simple and Modern identity look, do not mind the "middle name" parts, I forgot to remove them, they aren't included in this PR
image

### DIFF
--- a/[core]/esx_identity/html/css/style.css
+++ b/[core]/esx_identity/html/css/style.css
@@ -20,10 +20,10 @@ body {
     top: 50%;
     left: 50%;
     transform: translate(-50%, -50%);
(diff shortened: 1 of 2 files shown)


Gellipapa:
@nwvh  Hi! Thank you for this redesign. 

There are two errors. If you rewrite the identity, you have to write it in such a way that it is fully responsive. You should be able to choose from multiple themes, i.e. you should be able to manage themes that let the user specify which colours they want to use. 

In the HTML part you have included IDs for all input fields, but "ID" is included in the name of the field, it is a "unique identifier" you should use a class here.

PR #1319, opened 2024-02-19
chore(identity): added middle name feature

nwvh:
Adds option for players to choose a middle name (two first names) in the character registration UI


Gellipapa:
Hi! @nwvh Thanks, but here you need to better separate the code, so outsource the logic into a method and use it that way, by default this should be a configurable option and print is included in the code you opened for PR.

Gellipapa:
We can't accept this, you probably haven't even tested it because it includes the prints.

PR #1326, opened 2024-02-28
perf/refactor(client/functions): Refactor ESX.Game.GetClosestEntity for performance

itsmaty:
Refactored the ESX.Game.GetClosestEntity, instead of using two seperate k, v loops for filtering and distance checking i combined the logic into one faster for loop

also the coords parameter is either nil or an vector3 so if set, there is no need to parse it to an vector 3

i am not sure what the isPlayerEntities parameter is for and if true returning the index from the entities table (line 522), why?
i left that specific logic as is to not break anything

### DIFF
--- a/[core]/es_extended/client/functions.lua
+++ b/[core]/es_extended/client/functions.lua
@@ -496,29 +496,29 @@ function ESX.Game.IsSpawnPointClear(coords, maxDistance)
 end
 
 function ESX.Game.GetClosestEntity(entities, isPlayerEntities, coords, modelFilter)
-    local closestEntity, closestEntityDistance, filteredEntities = -1, -1, nil
+    local closestEntity, closestEntityDistance = -1, -1
 
     if not coords then
-        local playerPed = ESX.PlayerData.ped
-        coords = GetEntityCoords(playerPed)
+        local playerPed = ESX.PlayerData.ped;
+        coords = GetEntityCoords(playerPed);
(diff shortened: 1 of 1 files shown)


itsmaty:
so i made a mistake here
since the `isPlayerEntities` parameter is only used for `ESX.Game.GetClosestPlayer` because there `ESX.Game.GetPlayers ` is called like this `ESX.Game.GetPlayers(true, true)` (See Line 458) and that does that the ESX.Game.GetPlayers` function returns a key value table with the player as the key and the ped as the value

this gets passed into  `ESX.Game.GetClosestEntity` withe the `isPlayerEntities = true` and the ok, v loop would be indeed needed here because the loop schould return the player (the key of the table) and with an for loop we can only acess the value that what this line is for i woundered about
`closestEntity, closestEntityDistance = isPlayerEntities and k or entity, distance` in my refactor i replaced the k variable with the current loop index which is wrong.

I will do another commit, removing the logic for `isPlayerEntities` in `ESX.Game.GetClosestEntity` and calling `ESX.Game.GetPlayers(true)` so it will return an array containing the players. then the for loop in `ESX.Game.GetClosestEntity` will work an correctly return the player.

itsmaty:
ok nevermind you need the ped for distance but it should return the player so you need the player, ped k, v table
ok nothing i did made sende
just ignore

PR #1329, opened 2024-03-01
refactor(esx_core): spawning system, general code refactor

Thekuca:


bitpredator (approved):

Gellipapa:
@Thekuca Hi! We can't accept it like this, there are completely different codes, if refactor then refactor only, if feature then feature only. Don't put everything in one branch because it's hard to test and there's a big risk that the core won't work. This is very important to understand about core.

PR #1330, opened 2024-03-02
added heading again for xPlayer.getCoords()

TrymTube:
since it was removed, I added it again, which I did already in the past but somehow you missed the heading again


MoskalykA:
This will create compatibility problems if the script expects to receive a three-dimensional vector and not a four-dimensional vector ?

Gellipapa:
> This will create compatibility problems if the script expects to receive a three-dimensional vector and not a four-dimensional vector ?

It will work because the first 3 parameters of vector4 are xyz and so if you wrap in a vector3 then xyz will work as long as it doesn't have the heading value that w would have for vector4.

Gellipapa:
@TrymTube Hi! Thanks this improvement to getCoords.

PR #1327, opened 2024-02-28
refactor(client/functions): refactor getting burst tyres on vehicle

itsmaty:
i refactored the logic used to determine what tyres of an vehicles are burst

just get the wheel count and then iterate from 0 to the wheel count -1 and the current iteration is the tyre index

PR #1333, opened 2024-03-05
refactor(client/functions): refactor getting burst tyres on vehicle

itsmaty:
Refactored the code to determine the burst tyres of an vehicle

was already discussed in https://github.com/esx-framework/esx_core/pull/1327

@Gellipapa


Gellipapa:
@itsmaty Hi! Thanks i will test and merged if okey.

PR #1334, opened 2024-03-05
perf/refactor(client/functions): replace pairs loops with for loops where possible

itsmaty:
i replaced, where the functionality of pairs loops where not needed, these with regular for loops as there perform better

### DIFF
--- a/[core]/es_extended/client/functions.lua
+++ b/[core]/es_extended/client/functions.lua
@@ -507,10 +507,10 @@ function ESX.Game.GetClosestEntity(entities, isPlayerEntities, coords, modelFilt
 
     if modelFilter then
         filteredEntities = {}
-
-        for _, entity in pairs(entities) do
-            if modelFilter[GetEntityModel(entity)] then
-                filteredEntities[#filteredEntities + 1] = entity
+    
+        for current_entity_index = 1, #entities do
+            if modelFilter[GetEntityModel(entities[current_entity_index])] then
+                filteredEntities[#filteredEntities + 1] = entities[current_entity_index]
             end
         end
     end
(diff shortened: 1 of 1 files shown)


[core]/es_extended/client/functions.lua
```diff
@@ -507,10 +507,10 @@ function ESX.Game.GetClosestEntity(entities, isPlayerEntities, coords, modelFilt
 
     if modelFilter then
         filteredEntities = {}
-
-        for _, entity in pairs(entities) do
-            if modelFilter[GetEntityModel(entity)] then
-                filteredEntities[#filteredEntities + 1] = entity
+    
+        for current_entity_index = 1, #entities do
```
Comment (Gellipapa) [inline [core]/es_extended/client/functions.lua:511]: Please use snake_case instead of lowerCamelCase name convection.

[core]/es_extended/client/functions.lua
```diff
@@ -1128,8 +1128,8 @@ function ESX.ShowInventory()
                         { unselectable = true, icon = "fas fa-users", title = "Nearby Players" },
                     }
 
-                    for _, playerNearby in ipairs(playersNearby) do
-                        players[GetPlayerServerId(playerNearby)] = true
+                    for current_nearby_player_index = 1, #playersNearby do
```
Comment (Gellipapa) [inline [core]/es_extended/client/functions.lua:1131]: Please use snake_case instead of lowerCamelCase name convection.

Gellipapa (changes requested):

[core]/es_extended/client/functions.lua
Comment (itsmaty) [reply [core]/es_extended/client/functions.lua:511]: You mean lowerCamelCase instead of snake case right?

Comment (Gellipapa) [reply [core]/es_extended/client/functions.lua:511]: yes, exactly we don't use snake case in es_extended

Comment (itsmaty) [reply [core]/es_extended/client/functions.lua:511]: i will change it, probably tomorrow

itsmaty:
@Gellipapa 
chnaged

Gellipapa (approved):
Looks cool thanks.

PR #1338, opened 2024-03-27
refactor: use esx cache

MoskalykA:


Gellipapa:
Hi! It might be a bit misleading to say id, independent since the user would probably think of it as source, it should be called something else because it is an id but it is not the source, it would need wrapping GetPlayerServerId(PlayerId()) this is return player source.

[core]/es_extended/client/modules/actions.lua
```diff
@@ -28,13 +28,19 @@ CreateThread(function()
     while not ESX.PlayerLoaded do Wait(200) end
     while true do
         ESX.SetPlayerData("coords", GetEntityCoords(playerPed))
+        if playerId ~= PlayerId() then
+            playerId = PlayerId()
+
+            ESX.SetPlayerData("id", playerId)
```
Comment (tomiichx) [inline [core]/es_extended/client/modules/actions.lua:34]: I don't think this should be in a loop as the player id never changes - it is static. Same goes for the server id.

Comment (Gellipapa) [reply [core]/es_extended/client/modules/actions.lua:34]: @tomiichx Hi! Ohh yes nice catch it's true thanks.

mikigoalie:
playerid doesnt change but pedid changes... :D Just saying it's probably better to just get PlayerPedId() directly than caching in case of ped mismatch (either changed ped or whatever)

Gellipapa:
> playerid doesnt change but pedid changes... :D Just saying it's probably better to just get PlayerPedId() directly than caching in case of ped mismatch (either changed ped or whatever)

The PlayerPedId is already cached, but there's no point in caching the PlayerId, it's just more confusing if you enter the cached PlayerPedId and it doesn't work, so we won't merge it. 

Thanks for your understanding and comments.

PR #1340, opened 2024-04-02
Update legacy.sql

AmKinGesteR:
fix sql error


Gellipapa:
@AmKinGesteR Hi! Why duplicated insert job_grades table?

PR #1343, opened 2024-04-07
fix: ox_inventory setMaxWeight client-side sync

tomson701:
Client side maxWeight in ox_inventory is not synced when setting it using .Set


Gellipapa:
@tomson701 Hi! Thanks for fix sync weight in ox_inventory integration.

PR #1351, opened 2024-05-04
feat(Player): implement `esx:setGroup` event

tomiichx:


Gellipapa:
Hi! Thanks pr, can you explain some use cases when you would use this?

tomiichx:
> Hi! Thanks pr, can you explain some use cases when you would use this?

Hi!

I've had a few cases where I had to run loops to check for new admins or players who got their permissions set back to user. I usually cache all my results and then modify the table instead of fetching all players constantly. So one use case would be the admin chat, and a scoreboard for example. I have an admin count on the scoreboard, and if we add a new admin or remove one, I can't update my cached table without running a loop every minute or so...

I generally think this would be useful as `esx:setJob` is...

Gellipapa:
Hi! @tomiichx Thanks for this pr, i tested, it's works, merged.

PR #1358, opened 2024-05-28
fixed sql query main.lua

ethandudu:
Now the script can delete rows where the table name contains a special character like a -


Gellipapa:
@ethandudu Hi! Thanks, it's okey, merged.

PR #1359, opened 2024-06-01
fix(esx_menu_default): jquery handling element id

itIsMaku:
jQuery can't handle getting elements with `$("#...")` if the name contains a colon and takes it as another query. By getting the element via `document.getElementById` and wrapping it with `$(...)`, the functionality remains the same, but menu names can have a colon and other characters in them and the menu will open and not throw an error saying it didn't find the element.

Previously it was throwing `Uncaught TypeError: Cannot read properties of undefined (reading 'scrollIntoView') (@esx_menu_default/html/js/app.js:67)` because it couldn't find the element.

### DIFF
--- a/[core]/esx_menu_default/html/js/app.js
+++ b/[core]/esx_menu_default/html/js/app.js
@@ -62,7 +62,7 @@
         });
 
         ESX_MENU.render();
-        $("#menu_" + namespace + "_" + name)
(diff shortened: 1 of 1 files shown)


Gellipapa:
Hi! @itIsMaku Nice catch, it's not so nice to put the element in the selector in jquery, but it should be, after all it's a legacy system and mostly bug fixes are the priority, thanks for the fix.

PR #1365, opened 2024-06-16
fix: big bug

Arctos2win:


Who-is-Cola:
Have to delete the 'client/common.lua',  in es_extended\fxmanifest.lua too~~

PR #1373, opened 2024-06-17
fix(es_extended): removed heading from calculation

### LINKED ISSUE #1371: bug(es_extended): error's out on playerCoords
1️⃣ ISSUE
---
After last update, client errors out 
image

https://github.com/esx-framework/esx_core/blob/0bc232185aa05462d161638551f6bc7dbb7d7894/%5Bcore%5D/es_extended/client/main.lua#L484

Edit: looks like `playerCoords` is vec3 and `coords.pickup` is vec4

2️⃣ ISSUE
---
Also after last update, it's hard to pickup items, message pops up in server console:
`[WARNING] Player Detected Cheating (Out of range pickup)`
I think since there was some changes with coords, heading is being saved too and it's getting calculated towards coordinates.

---

⚠️ _Using default inventory (no ox_inventory)_

## EDIT: pushed 2 updates for these issues.

iSentrie:
Fix #1371 part 2

### DIFF
--- a/[core]/es_extended/server/main.lua
+++ b/[core]/es_extended/server/main.lua
@@ -546,7 +546,7 @@ if not Config.OxInventory then
         local pickup, xPlayer, success = Core.Pickups[pickupId], ESX.GetPlayerFromId(source)
 
         if pickup then
-            local playerPickupDistance = #(pickup.coords - xPlayer.getCoords(true))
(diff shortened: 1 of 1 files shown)


Arctos2win:
> Fix #1371 part 2

Hi @iSentrie, 

Thanks for the fix but i think this specific problem needs solving at the function. 
Since the update that was done to the function will not only break this but also other things. 

Greetings 
- @Arctos2win

PR #1381, opened 2024-07-15
allowing the player to use the old ways to get the shared object while in the same time allowing the export

MohamexDR:
# as the title says
## allows the player to use the old ways to get the shared object while in the same time allowing the export

### DIFF
--- a/[core]/cron/fxmanifest.lua
+++ b/[core]/cron/fxmanifest.lua
@@ -4,6 +4,6 @@ game 'gta5'
 author 'ESX-Framework'
 description 'Allows resources to Run tasks at specific intervals.'
 lua54 'yes'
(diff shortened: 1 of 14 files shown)


Gellipapa:
@MohamexDR Hi! Many people have already switched to the new version and can use it, but I understand your concern and given that it's a legacy system and it was previously released after a bad rash decision, we will reconsider that in case we see activity that you want it back. 

Please report it with a github emoji if you want getSharedObject back.

Thank you in advance for your understanding.

Gellipapa:
https://github.com/esx-framework/esx_core/pull/933/files/5b67a38fe7899a720aadbc3157a6f0e657a14d5e

PR #1384, opened 2024-07-25
fix(server): attempt to index a nil value (local 'xPlayer')

someone005:
This error occurs when you use command that doesn't exist and your data is not loaded (command used in F8 while I was in multicharacter menu)


TheFantomas (approved):

Gellipapa:
@someone005 Hi! Thanks for fixes little bug.

PR #1403, opened 2024-09-23
Typo fix

STACHY225:
The actual argument (job name) from .setJob() will be printed, not some `table:ABCDEFG` from `CreateExtendedPlayer`.


Arctos2win:
Hi @STACHY225, 

thanks for fixing this typo🙂

PR #1404, opened 2024-09-27
Dev - Fix job creation with full transaction for `jobs` and `job_grades` tables

Mesrine67:
This patch improves the job creation process by using a complete SQL transaction to ensure that insertions into both the `jobs` and `job_grades` tables are atomic. The use of `INSERT IGNORE INTO jobs` has been corrected to avoid inconsistencies in retrieving the job ID, and all insertion steps are now part of a transaction to ensure that the job and its grades are created reliably.

image

```lua
lib.callback.register('advanced_management:staff:creator', function(source, type, data)
    if type == 'job' then
        local name = data.name
        local label = data.label
        local grades = data.grades
        local success = ESX.CreateJob(name, label, grades)
        -- Wait(500)
        if not success then
            return false
        end
        local jobs = ESX.GetJobs()
        if jobs[name] then
            lib.notify(source, {title = 'Success', description = 'The job was successfully created.', type = 'success'})
            return true
        else
            lib.notify(source, {title = 'Error', description = 'Job creation failed.', type = 'error'})
            return false
        end
    end
    return false
end)
```

### DIFF
--- a/[core]/es_extended/server/modules/createJob.lua
+++ b/[core]/es_extended/server/modules/createJob.lua
@@ -19,18 +19,6 @@ local function doesJobAndGradesExist(name, grades)
    return true
 end
 
-local function generateTransactionQueries(name,grades)
-    local queries = {}
-    for _, grade in ipairs(grades) do
-        queries[#queries+1] = {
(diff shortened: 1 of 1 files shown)


Gellipapa (approved):
Hi! Thanks for cleanup, and refactor this code block.

PR #1406, opened 2024-10-01
fix(notify): no valid icon

someone005:
If invalid notify type is provided, it throws an error to console


Gellipapa (approved):

Gellipapa:
@someone005 Hi! Thanks for handle invalid icon.

someone005:
> @someone005 Hi! Thanks for handle invalid icon.

No problem :)

PR #1407, opened 2024-10-08
!Refactor(core): Statebag removal and changes

Mycroft-Studios:
Currently, everything flowing through xPlayer.set is being put into a statebag, aswell as all metadata , resulting in crashes.

This change is a high priority.

### DIFF
--- a/[core]/es_extended/client/main.lua
+++ b/[core]/es_extended/client/main.lua
@@ -708,7 +708,3 @@ end)
 ESX.RegisterClientCallback("esx:GetVehicleType", function(cb, model)
     cb(ESX.GetVehicleType(model))
 end)
-
-AddStateBagChangeHandler("metadata", "player:" .. tostring(GetPlayerServerId(PlayerId())), function(_, key, val)
-    ESX.SetPlayerData(key, val)
-end)
--- a/[core]/es_extended/server/classes/player.lua
+++ b/[core]/es_extended/server/classes/player.lua
@@ -52,12 +52,11 @@ function CreateExtendedPlayer(playerId, identifier, group, accounts, inventory, 
     _ExecuteCommand(("add_principal identifier.%s group.%s"):format(self.license, self.group))
 
     local stateBag = Player(self.source).state
-    stateBag:set("identifier", self.identifier, true)
-    stateBag:set("license", self.license, true)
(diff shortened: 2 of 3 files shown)


thelindat:
please fix

Gellipapa:
@Mycroft-Studios Hi! Thanks for reporting the problem we will fix it. However, I have two questions: 

Is the statebag crash because of the continuous data setup or because there is a lot of data? 

I couldn't find any information anywhere that the statebag is only designed for a few tables of data. 

I think that it crashes because it is setting the data continuously, so it's not the amount of statebag data but the frequency that is the problem. 

How do you see this?

thelindat:
Statebags are not intended for _large_ datasets. They should be kept as small as possible, with larger sets of data being broken up into multiple states _if possible and absolutely required_. Furthermore, statebags are synced to _all_ clients - any private data being stored with xPlayer.set or xPlayer.setMeta is synced to everyone. This sync occurs when the data is set _and_ when entering the player's scope.

Random unstructured unknown infinitely growing data shouldn't be set into statebags. Period. Having xPlayer.set synced is bad enough, but all metadata being on a single statebag is horrendous.

Gellipapa:
> Statebags are not intended for _large_ datasets. They should be kept as small as possible, with larger sets of data being broken up into multiple states _if possible and absolutely required_. Furthermore, statebags are synced to _all_ clients - any private data being stored with xPlayer.set or xPlayer.setMeta is synced to everyone. This sync occurs when the data is set _and_ when entering the player's scope.
> 
> Random unstructured unknown infinitely growing data shouldn't be set into statebags. Period. Having xPlayer.set synced is bad enough, but all metadata being on a single statebag is horrendous.

Hi! I understand and accept that this was a bad implementation. Can you say the same about globalstate? Because as I looked at the implementation the statebag seems to me more stable than the GlobalState bag. What is your opinion about this?

thelindat:
Large data is still not good for GlobalState, and statebags will just stop syncing once they hit a certain size.

Granular updates are not possible with statebags either, so either splitting statebags up logically or using events for updating certain key-values is preferable.

Gellipapa (approved):

Gellipapa:
@Mycroft-Studios @Arctos2win Thanks for fixed this urgent problem.

PR #1410, opened 2024-10-12
Add an option to use the cancel function instead of the close function when closing a menu or all menus.

zImSkillz:
Should be self-explanatory, it would make things a lot easier if you could close directly with the cancel function instead of just using the close function.

### DIFF
--- a/[core]/es_extended/client/functions.lua
+++ b/[core]/es_extended/client/functions.lua
@@ -236,21 +236,29 @@ function ESX.UI.Menu.Open(menuType, namespace, name, data, submit, cancel, chang
     return menu
 end
 
-function ESX.UI.Menu.Close(menuType, namespace, name)
+function ESX.UI.Menu.Close(menuType, namespace, name, cancel)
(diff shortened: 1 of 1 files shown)


Gellipapa:
@zImSkillz Hi! If I understand you correctly you mean that cancel is included in self but not used so you need to specify an extra parameter to write a close callback so you want the cancel method to run close if you specify it? Am I understanding this correctly?

zImSkillz:
> @zImSkillz Hi! If I understand you correctly you mean that cancel is included in self but not used so you need to specify an extra parameter to write a close callback so you want the cancel method to run close if you specify it? Am I understanding this correctly?

Something like this, the Menu Close and Close All function only triggers the close function from the menu, and what I would like is that you can use the cancel function per argument instead of the close function, as already implemented in my commit, so you save yourself some code having to include the same cancel function in close as well.

Gellipapa:
Yes I thought so, thanks for the improvement.

PR #1411, opened 2024-10-13
fix(es_extended/client/functions.lua) RegisterInput and HashString bug

feelfreetofee:
In the last update, **RegisterInput** tries to execute the command on key release due some checks removal in #1303. The check was restored and tested.

**HashString** breaks when the joaat returns a hash bigger than 2^31. It was fix using a [shift mask](https://stackoverflow.com/a/20294714).

### DIFF
--- a/[core]/es_extended/client/functions.lua
+++ b/[core]/es_extended/client/functions.lua
@@ -120,7 +120,7 @@ function ESX.DrawMissionText(msg, time)
 end
 
 function ESX.HashString(str)
-    return ('~INPUT_%s~'):format(('%x'):format(joaat(str)):upper())
+    return ('~INPUT_%s~'):format(('%x'):format(joaat(str) & 0x7fffffff + 2 ^ 31):upper())
 end
 
 function ESX.OpenContext(...)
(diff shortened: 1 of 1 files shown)


[core]/es_extended/client/functions.lua
```diff
@@ -140,12 +140,12 @@ function ESX.RefreshContext(...)
...
-    RegisterCommand("+" .. command_name, on_press)
-    Core.Input[command_name] = ESX.HashString("+" .. command_name)
+    RegisterCommand(on_release and '+' .. command_name or command_name, on_press)
+    Core.Input[command_name] = ESX.HashString(command_name)
     if on_release then
-        RegisterCommand("-" .. command_name, on_release)
+        RegisterCommand('-' .. command_name, on_release)
     end
-    RegisterKeyMapping("+" .. command_name, label or '', input_group or 'keyboard', key or '')
+    RegisterKeyMapping(command_name, label or '', input_group or 'keyboard', key or '')
```
Comment (Gellipapa) [inline [core]/es_extended/client/functions.lua:148]: Why removed '+' if you used 143 row?

Gellipapa (changes requested):

[core]/es_extended/client/functions.lua
Comment (feelfreetofee) [reply [core]/es_extended/client/functions.lua:148]: Made some testing, and apparently the plus sign is not required for **RegisterKeyMapping**. Tested in both cases where on_release is present and not.
Identical case for **HashString** inside of **RegisterInput**.
The only case where FiveM tries to trigger **on_release** is when the command name in **RegisterCommand** starts with a plus sign (+).
When **on_release** is not present, the command name in **RegisterCommand** doesn't use the symbol to prevent the following bug
image

Comment (Gellipapa) [reply [core]/es_extended/client/functions.lua:148]: Thanks for test, and research, i merged.

Gellipapa (approved):

Tgrp:
Thank you very much, I had this problem too

PR #1416, opened 2024-10-15
tweak: use txAdmin locale by default 

Mycroft-Studios:
by default, use the txAdmin locale, unless a `esx:locale` is specified.
This allows people to change languages easier, since everything will be in the same place, and if they want to specify a different language for ESX, then they can do.

This also wont break anything, since it wont affect existing users whom will aready have the `esx:locale` set

### DIFF
--- a/[core]/es_extended/config.lua
+++ b/[core]/es_extended/config.lua
@@ -3,7 +3,7 @@ Config = {}
 local txAdminLocale = GetConvar("txAdmin-locale", "en")
 local esxLocale = GetConvar("esx:locale", "none")
 
-Config.Locale = esxLocale == "none" and txAdminLocale or esxLocale
+Config.Locale = (esxLocale == "none" and txAdminLocale ~= "custom") and txAdminLocale or esxLocale
 
 Config.OxInventory = GetResourceState("ox_inventory") ~= 'missing'
 
--- a/[core]/es_extended/config.lua
+++ b/[core]/es_extended/config.lua
@@ -1,13 +1,9 @@
 Config = {}
 
 local txAdminLocale = GetConvar("txAdmin-locale", "en")
-local esxLocale = GetConvar("esx:locale", "none")
+local esxLocale = GetConvar("esx:locale", "invalid")
 
-if esxLocale == "none" then
-    Config.Locale = txAdminLocale ~= "custom" and txAdminLocale or "en"
-else
-    Config.Locale = esxLocale
-end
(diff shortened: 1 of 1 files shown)


Gellipapa:
@Mycroft-Studios Hi! 

Why not make it so that you include the get-er of txadmin in the esx locale, if it is given before then it will use it anyway if not given then txadmin will use it and if txadmin is not given then say a default en and then you don't need the extra variables and additional conditions.

Mycroft-Studios:
wdym by that?

Gellipapa:
Something like that.
If there is no esx:locale it will check if there is something in txadmin if there is it will set it if not it will use 'en'.

`Config.Locale = GetConvar("esx:locale", GetConvar("txAdmin-locale", "en"))`

Mycroft-Studios:
> Something like that. If there is no esx:locale it will check if there is something in txadmin if there is it will set it if not it will use 'en'.
> 
> `Config.Locale = GetConvar("esx:locale", GetConvar("txAdmin-locale", "en"))`

there is a chance the txAdmin-locale = "custom", and therefore that needs to be taken into account

Gellipapa:
> > Something like that. If there is no esx:locale it will check if there is something in txadmin if there is it will set it if not it will use 'en'.
> > `Config.Locale = GetConvar("esx:locale", GetConvar("txAdmin-locale", "en"))`
> 
> there is a chance the txAdmin-locale = "custom", and therefore that needs to be taken into account

Hm, and what language is custom? Could this be a design flaw? I think you should also specify "en" instead of custom in default, I don't understand what the custom value is in txadmin.

`Config.Locale = esxLocale ~= "none" and esxLocale or (txAdminLocale ~= "custom" and txAdminLocale or "en")`

Mycroft-Studios:
People love todo things the wrong way, so custom is there for to stop people breaking stuff if they edit the tx translations in the wrong way.

And while the one liner would work, i feel like its.. less readable

Gellipapa:
> People love todo things the wrong way, so custom is there for to stop people breaking stuff if they edit the tx translations in the wrong way.
> 
> And while the one liner would work, i feel like its.. less readable

I understand, but if something is not ready, why is it not easier to set the language to "en" as default, rather than adding a custom type which would result in "en" anyway. 

Yes there are cases where it's not readable if you shortcut, but personally this is an easy readable version and I find it unnecessary to make more conditonal.

Gellipapa (approved):

PR #1430, opened 2024-10-28
Adding a option to disable a paycheck for a specific user

zImSkillz:
# Explanation:
Adding an option to prevent a specific user from receiving their paychecks

# Why?
Since some servers have the function to disable the paycheck for a user if the user is AFK for more than ... minutes/hours.. , and since this function is not implemented directly in ESX, it has to be added manually, which becomes annoying over time because you always have to add the function again when you update ESX, so it would make sense to simply add a function that allows this

### DIFF
--- a/[core]/es_extended/server/paycheck.lua
+++ b/[core]/es_extended/server/paycheck.lua
@@ -1,8 +1,10 @@
 function StartPayCheck()
     local disabledPaycheckUsers = {}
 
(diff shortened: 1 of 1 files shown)


Mycroft-Studios:
added with [f780e84](https://github.com/esx-framework/esx_core/commit/f780e843936cf745b4606b4d6e9ba86dcc2671e0)

PR #1433, opened 2024-10-31
feat(esx_core): ✨ make adding other inventories more easy + little refactor

Arctos2win:


Kenshiin13:
In my opinion, it would have been more appropriate to split this into two PRs.

PR #1436, opened 2024-11-02
refactor(SQL/legacy): Add primary key to vehicles table

Kenshiin13:
### Description:
This PR sets the `model` column as the primary key for the existing `vehicles` table.

---

### Motivation:
- **Data Integrity**: The primary key enforces uniqueness, ensuring each vehicle can be distinctly identified.
- **Ease of Editing**: With a primary key in place, we can easily manage and edit rows in tools like phpMyAdmin.

### DIFF
--- a/[SQL]/legacy.sql
+++ b/[SQL]/legacy.sql
@@ -427,12 +427,10 @@ CREATE TABLE `user_licenses` (
 --
 
 CREATE TABLE `vehicles` (
-  `id` int(11) NOT NULL AUTO_INCREMENT,
   `name` varchar(60) NOT NULL,
   `model` varchar(60) NOT NULL,
   `price` int(11) NOT NULL,
-  `category` varchar(60) DEFAULT NULL,
-  PRIMARY KEY (`id`)
+  `category` varchar(60) DEFAULT NULL
 ) ENGINE=InnoDB;
 
 --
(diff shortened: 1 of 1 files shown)


Rav3n95:
If something i would add the primary key to the model instead of a new id columb. The reason behind it, its because you can still make duplicates.

Kenshiin13:
> If something i would add the primary key to the model instead of a new id columb. The reason behind it, its because you can still make duplicates.

Which may be intended. I could imagine having the same vehicle model in 2 different categories with a different price.

Bossman02:
That makes no sense, why do you want to have the same vehicle in 2 categories with different prices?, nobody is going to buy the one in the expensive category.

I agree with raven, it is the change I always make to this table

Kenshiin13:
> That makes no sense, why do you want to have the same vehicle in 2 categories with different prices?, nobody is going to buy the one in the expensive category.
> 
> I agree with raven, it is the change I always make to this table

Nevermind, I just checked the vehicleshop script again and I have to agree with both of you. I thought you were able to limit access to certain categories to specific groups/jobs out of the box, but that is not the case. So there really is no reason to allow for duplicate models.

Kenshiin13:
@Arctos2win @Mycroft-Studios

Arctos2win (approved):

PR #1460, opened 2024-11-13
General Improvement

Mycroft-Studios:


Kenshiin13 (dismissed):
It's not Friday, that's good enough for me.

Kenshiin13 (changes requested):
image

Error when creating a character on a fresh install

PR #1461, opened 2024-11-13
DOB validation removal from ESX_identity

luckie12:
Removed DOB validation as the weird dob notation from overseas doesnt make sense/work


Kenshiin13:
@luckie12 @Mycroft-Studios  These changes were not needed and didn't fix the problem at hand. I've reverted all commits and pushed an actual fix (hopefully).

Mycroft-Studios (approved):

PR #1472, opened 2024-11-16
fix(es_extended/client/functions.lua): bug fix with the menu close & closeall function

zImSkillz:
As the title says, this fixes a bug that occurs when using the menu- close or close all function with the cancel argument set.

The error was that the Cancel function was called without any further arguments, as it normally is.


Kenshiin13 (changes requested):
Please stick to the [PR Guidelines](https://github.com/esx-framework/esx_core/issues/343).

Kenshiin13 (approved):

Kenshiin13 (changes requested):

Kenshiin13 (approved):
LGTM, thanks!

PR #1481, opened 2024-11-17
feat(es_extended/client/callback): implement ESX.AwaitServerCallback

Mycroft-Studios:
example usage:
```lua
RegisterCommand("callback", function ()
    local value, value2 = ESX.AwaitServerCallback("es_extended:callbackTest")
    print(value, value2)
end, false)
```

### DIFF
--- a/[core]/es_extended/client/modules/callback.lua
+++ b/[core]/es_extended/client/modules/callback.lua
@@ -41,7 +41,7 @@ function Callbacks:ServerRecieve(requestId, invoker, ...)
     local callback = self.requests[self.currentId]
 
     if callback.await then
-        return callback.cb:resolve({...})
+        callback.cb:resolve({...})
     else
         self:Execute(callback.cb, ...)
     end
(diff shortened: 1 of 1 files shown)


[core]/es_extended/client/modules/callback.lua
```diff
@@ -32,7 +40,12 @@ function Callbacks:ServerRecieve(requestId, invoker, ...)
...
     local callback = self.requests[self.currentId]
 
-    Callbacks:Execute(callback, ...)
+    if callback.await then
+        return callback.cb:resolve({...})
+    else
+        Callbacks:Execute(callback.cb, ...)
+    end
+
     self.requests[requestId] = nil
```
Comment (Kenshiin13) [inline [core]/es_extended/client/modules/callback.lua:49]: We should probably do this before executing the callback to avoid stale entries in the requests table if the callback throws an error.

Kenshiin13 (changes requested):

Kenshiin13 (approved):
LGTM!

PR #1503, opened 2024-11-24
Spawn vehicle in right routing bucket

Arius-Scripts:
# Fix Routing Bucket Issue in /car command
## Description
This pull request resolves an issue where a vehicle spawned for a player in a different routing bucket (e.g., when the player is in a non-default bucket) results in the error:  
**"The player could not be seated in the vehicle".**  

The issue occurs because the spawned vehicle is created in the default bucket (bucket 0) while the player resides in another bucket.  

---

## Changes Made
- Fetching the player's current routing bucket using `GetPlayerRoutingBucket()`.
- Set the routing bucket of the spawned vehicle to match the player's routing bucket using `SetEntityRoutingBucket()`.


Mycroft-Studios:
This is already fixed in dev

PR #1505, opened 2024-11-25
feat(es_extended): Extra callback functions

Mycroft-Studios:
Introduces 2 new functions:
`ESX.DoesClientCallbackExist`
`ESX.DoesServerCallbackExist`


Kenshiin13:
I don't see the point of this. ESX already throws an error if a resource attempts to access an invalid callback. This just promotes people calling `ESX.DoesServerCallbackExist` and `ESX.DoesClientCallbackExist` before every damn callback.

Third party resources can check if their dependencies are started and just assume that all expected callbacks are registered. We are already letting them know if they are not ^^

Arctos2win:
> I don't see the point of this. ESX already throws an error if a resource attempts to access an invalid callback. This just promotes people calling `ESX.DoesServerCallbackExist` and `ESX.DoesClientCallbackExist` before every damn callback.
> 
> Third party resources can check if their dependencies are started and just assume that all expected callbacks are registered. We are already letting them know if they are not ^^

I agree with Kenshin on this one, there is 0 usecase for it and litterly no one ever requested a functionality like this 
in the so many years this project exists.

[core]/es_extended/client/modules/callback.lua
```diff
@@ -98,17 +115,21 @@ ESX.AwaitServerCallback = function(eventName, ...)
...
+---@return boolean
+function ESX.DoesClientCallbackExist(eventName)
+    return Callbacks.storage[eventName] ~= nil
+end
+
+---@param eventName string
+---@return table | nil
+function ESX.GetClientCallbackInfo(eventName)
+    return Callbacks.storage[eventName]
+end
```
Comment (Kenshiin13) [inline [core]/es_extended/client/modules/callback.lua:135]: This doesn't need to be a thing

[core]/es_extended/server/modules/callback.lua
```diff
@@ -80,7 +80,18 @@ end
...
+---@return boolean
+function ESX.DoesServerCallbackExist(eventName)
+    return Callbacks.storage[eventName] ~= nil
+end
+
+---@param eventName string
+---@return table | nil
+function ESX.GetServerCallbackInfo(eventName)
+    return Callbacks.storage[eventName]
+end
```
Comment (Kenshiin13) [inline [core]/es_extended/server/modules/callback.lua:97]: This doesn't need to be a thing

Kenshiin13 (changes requested):

Kenshiin13:
@Mycroft-Studios

Kenshiin13 (approved):
LGTM!

PR #1506, opened 2024-11-25
String error

HydraDevFR:
SCRIPT ERROR: @esx_multicharacter/server/modules/database.lua:22: attempt to perform arithmetic on a nil value

I noticed that when using set mysql_connection_string "server=, it didn't work. However, using set mysql_connection_string "mysql:// worked. I have adapted the code so that it supports both versions.

### DIFF
--- a/[core]/esx_multicharacter/server/modules/database.lua
+++ b/[core]/esx_multicharacter/server/modules/database.lua
@@ -10,20 +10,27 @@ function Database:GetConnection()
         error(connectionString .. "\n^1Unable to start Multicharacter - unable to determine database from mysql_connection_string^0", 0)
     elseif connectionString:find("mysql://") then
         connectionString = connectionString:sub(9, -1)
-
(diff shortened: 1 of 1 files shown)


Mycroft-Studios:
(closed for 2 reaons: 1. user error, 2. that method is from mysql-async, which hasnt been used since ESX 1.3, years ago,
that code is *only* in esx multicharacter, because they were made around the same time, and was there for users upgrading from mysql-async. however its 2024, theres 0 reason to not have the correct version in your server cfg)

HydraDevFR:
Hello, I wanted to follow up with you regarding this issue. This was tested
in an environment using esx_core <https://github.com/esx-framework/esx_core>
and the latest version of oxmysql.

I want to clarify that I always pay close attention to the versions I use.
In this specific case, I had just performed a test for a client. However,
while trying to set up a fresh environment, I encountered this issue.

Le lun. 25 nov. 2024 à 23:32, Kasey Fitton ***@***.***> a
écrit :

PR #1513, opened 2024-11-27
feat(es_extended): Add Entity Lockdown by default

Mycroft-Studios:


Kenshiin13 (approved):
LGTM

Kenshiin13:
We should probably make sure that ESX is compatible with strict mode. Specifically props spawned through esx addons could cause issues.

PR #1515, opened 2024-11-29
fix(esx_multicharacter/server/modules/database): improve database name extraction

RrybaN:
This fix resolves an issue where parsing the `mysql_connection_string` in the format `user=root;password=12345;host=localhost;port=3306;database=fivem` would throw an error `attempt to perform arithmetic on a nil value`. Additionally, i added a error log if database name is not found in the `mysql_connection_string`


RrybaN:
@Mycroft-Studios any info why this was closed?

Mycroft-Studios:
sinced since #1516  Is the correct fix.
This part of the code, was wrote by ox, so i get the feeling their extract method is alot more accuate :)

RrybaN:
Hmm, well okay, got it.

PR #1516, opened 2024-11-29
fix(esx_multicharacter/server): wrong database syntax

Mycroft-Studios:


Kenshiin13 (changes requested):
We've talked about getting rid of this entirely

PR #1517, opened 2024-11-29
chore(es_extended/server): add missing identifier config

RrybaN:
I noticed that the Config.Identifier was being used in esx_multicharacter but not existing on es_extended config so i added it.

https://github.com/esx-framework/esx_core/blob/efafa58eb2e24c3bd0464d50ae86f9639da52b67/%5Bcore%5D/esx_multicharacter/server/main.lua#L7

### DIFF
--- a/[core]/es_extended/server/functions.lua
+++ b/[core]/es_extended/server/functions.lua
@@ -368,8 +368,8 @@ function ESX.GetIdentifier(playerId)
 
     playerId = tostring(playerId)
 
-    local identifier = GetPlayerIdentifierByType(playerId, "license")
-    return identifier and identifier:gsub("license:", "")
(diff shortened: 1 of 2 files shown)


Mycroft-Studios:
Using anything other than licence is extremely stupid, and a great way to have people exploit your server.
IP: I dont even need to explain this.. i hope
Discord: can create an alt account within minutes and abuse the system,
Steam: can create an alt account within minutes and abuse the system,
Licence2: not all users have 1

RrybaN:
I know that players can create alt accounts, and it's ultimately up to the server owner whether they want to use different licenses. For example, if you get a new PC or reformat your own PC, sometimes your license will change, and of course, you cannot retrieve your character. So, it's better to use Steam or Discord licenses in that case.

PR #1532, opened 2024-12-08
refactor: Convar to use FiveM identifier as main

Mycroft-Studios:


Arctos2win (changes requested):
Maybe if we are doing it like this its better if we give the full choice between 
license, fivemid and steamid

FwbStudio:
in extended can be Config.MainIdentifier but it should have rockstarlic and fivem lic atleast

Mycroft-Studios:
> in extended can be Config.MainIdentifier but it should have rockstarlic and fivem lic atleast

This is a change that will be happening in a few *months* , not days, and was consulted with Cfx on

PR #1540, opened 2024-12-15
feat(es_extended) Add ESX.Await function

Zykem:
This PR adds a utility function, ESX.Await, to simplify waiting for a condition without the need for manual loops. It checks a condition repeatedly until it’s met or the timeout is reached.

Example usage:

```lua
local networkId = ESX.AwaitServerCallback('spawnVehicle')

local function awaitExistence()
    return NetworkDoesEntityExistWithNetworkId(networkId)
end

-- timeElapsed value in MS
local success, timeElapsed = ESX.Await(awaitExistence, 'Timed out :(', 3000)

-- we can now mess with the vehicle spawned server-side on the client-side
```

### DIFF
--- a/[core]/es_extended/shared/functions.lua
+++ b/[core]/es_extended/shared/functions.lua
@@ -172,46 +172,40 @@ function ESX.IsFunctionReference(val)
     return typeVal == "function" or (typeVal == "table" and type(getmetatable(val)?.__call) == "function")
 end
 
----@param conditionFunc function A function that returns a boolean indicating whether the condition is met.
----@param errorMessage? string The error message to print if the condition is not met within the timeout period.
----@param timeout? number The maximum time (in milliseconds) to wait for the condition to be met.
----@return boolean, number: Returns success status and the time taken
-ESX.Await = function(conditionFunc, errorMessage, timeout)
-    timeout = timeout or 1000
-    
-    if timeout < 0 then
-        error('Timeout should be a positive number.')
+---@param conditionFunc function A function that is repeatedly called until it returns a truthy value or the timeout is exceeded.
+---@param errorMessage? string Optional. If set, an error will be thrown with this message if the condition is not met within the timeout. If not set, no error will be thrown.
+---@param timeoutMs? number Optional. The maximum time to wait (in milliseconds) for the condition to be met. Defaults to 1000ms.
+---@return boolean, any: Returns success status and the returned value of the condition function.
+function ESX.Await(conditionFunc, errorMessage, timeoutMs)
+    timeoutMs = timeoutMs or 1000
+
+    if timeoutMs < 0 then
+        error("Timeout should be a positive number.")
     end
 
-    local isFunctionReference = ESX.IsFunctionReference(conditionFunc)
-
-    if not isFunctionReference then
-        error('Condition Function should be a function reference.')
+    if not ESX.IsFunctionReference(conditionFunc) then
+        error("Condition Function should be a function reference.")
     end
 
     -- since errorMessage is optional, we only validate it if the user provided it.
     if errorMessage then
-        ESX.AssertType(errorMessage, 'string', 'errorMessage should be a string.')
+        ESX.AssertType(errorMessage, "string", "errorMessage should be a string.")
     end
 
-    local startTime = GetGameTimer()
-    local prefix = ('[%s] -> '):format(GetInvokingResource())
-
-    while GetGameTimer() - startTime < timeout do
+    local startTimeMs = GetGameTimer()
+    while GetGameTimer() - startTimeMs < timeoutMs do
         local result = conditionFunc()
 
         if result then
-            local elapsedTime = GetGameTimer() - startTime
-            return true, elapsedTime
+            return true, result
         end
 
         Wait(0)
     end
 
     if errorMessage then
-        local formattedErrorMessage = ('%s %s'):format(prefix, errorMessage)
-        error(formattedErrorMessage)
+        error(("[%s] -> %s"):format(GetInvokingResource(), errorMessage))
     end
 
-    return false, timeout
+    return false
 end
--- a/[core]/es_extended/shared/functions.lua
+++ b/[core]/es_extended/shared/functions.lua
@@ -192,6 +192,7 @@ function ESX.Await(conditionFunc, errorMessage, timeoutMs)
         ESX.AssertType(errorMessage, "string", "errorMessage should be a string.")
     end
 
+    local invokingResource = GetInvokingResource()
     local startTimeMs = GetGameTimer()
     while GetGameTimer() - startTimeMs < timeoutMs do
         local result = conditionFunc()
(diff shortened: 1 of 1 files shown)


[core]/es_extended/shared/functions.lua
```diff
@@ -170,4 +170,43 @@ function ESX.IsFunctionReference(val)
     local typeVal = type(val)
 
     return typeVal == "function" or (typeVal == "table" and type(getmetatable(val)?.__call) == "function")
+end
+
+---@param conditionFunc function A function that returns a boolean indicating whether the condition is met.
+---@param timeout number The maximum time (in milliseconds) to wait for the condition to be met.
+---@param errorMessage? string The error message to print if the condition is not met within the timeout period.
+---@return boolean: Returns true if the condition is met within the timeout, otherwise returns `false`.
+---@return number: The time (in milliseconds) it took to meet the condition, or the timeout time if not met.
```
Comment (Kenshiin13) [inline [core]/es_extended/shared/functions.lua:179]: Returns boolean, number.

[core]/es_extended/shared/functions.lua
```diff
@@ -170,4 +170,43 @@ function ESX.IsFunctionReference(val)
     local typeVal = type(val)
 
     return typeVal == "function" or (typeVal == "table" and type(getmetatable(val)?.__call) == "function")
+end
+
+---@param conditionFunc function A function that returns a boolean indicating whether the condition is met.
+---@param timeout number The maximum time (in milliseconds) to wait for the condition to be met.
```
Comment (Kenshiin13) [inline [core]/es_extended/shared/functions.lua:176]: Timeout is optional.

[core]/es_extended/shared/functions.lua
```diff
@@ -170,4 +170,43 @@ function ESX.IsFunctionReference(val)
...
+---@param errorMessage? string The error message to print if the condition is not met within the timeout period.
+---@return boolean: Returns true if the condition is met within the timeout, otherwise returns `false`.
+---@return number: The time (in milliseconds) it took to meet the condition, or the timeout time if not met.
+ESX.Await = function(conditionFunc, timeout, errorMessage)
```
Comment (Kenshiin13) [inline [core]/es_extended/shared/functions.lua:180]: Use `ESX.IsFunctionReference` to validate conditionFunc.

Personally, I'd also prefer to have `errorMessage` as the second param. More often than not, we will want to provide an error message without increasing the default timeout. Makes it a little easier to work with this function.

[core]/es_extended/shared/functions.lua
```diff
@@ -170,4 +170,43 @@ function ESX.IsFunctionReference(val)
...
+            local elapsedTime = GetGameTimer() - startTime
+            return false, elapsedTime
+        end
+
+        if result then
+            local elapsedTime = GetGameTimer() - startTime
+            return true, elapsedTime
+        end
+
+        Wait(10)
```
Comment (Kenshiin13) [inline [core]/es_extended/shared/functions.lua:204]: Any particular reason we are not running this every frame? This just seems to limit what we can do with this function.

[core]/es_extended/shared/functions.lua
```diff
@@ -170,4 +170,43 @@ function ESX.IsFunctionReference(val)
...
+    end
+
+    local startTime = GetGameTimer()
+    local prefix = ('[%s] -> '):format(GetInvokingResource())
+
+    while GetGameTimer() - startTime < timeout do
+        local success, result = pcall(conditionFunc)
+
+        if not success then
+            print(prefix .. 'Error while calling conditionFunc! Result: ' .. result)
```
Comment (Kenshiin13) [inline [core]/es_extended/shared/functions.lua:194]: Can we stick to string formatting, please.

[core]/es_extended/shared/functions.lua
```diff
@@ -170,4 +170,43 @@ function ESX.IsFunctionReference(val)
...
+        if result then
+            local elapsedTime = GetGameTimer() - startTime
+            return true, elapsedTime
+        end
+
+        Wait(10)
+    end
+
+    if errorMessage then
+        print(prefix .. errorMessage)
```
Comment (Kenshiin13) [inline [core]/es_extended/shared/functions.lua:208]: Can we stick to string formatting, please.

Kenshiin13 (changes requested):

[core]/es_extended/shared/functions.lua
Comment (Zykem) [reply [core]/es_extended/shared/functions.lua:204]: This was only made for "optimizaiton" purposes, i did not see any needs for running this every frame. Although now that you say this, i will refactor this to run every frame to cover every use-case.

[core]/es_extended/shared/functions.lua
Comment (Zykem) [reply [core]/es_extended/shared/functions.lua:180]: Do you think the function should provide a default error message, or just not print anything if it doesnt hit the condition within the `timeout` param?

Comment (Kenshiin13) [reply [core]/es_extended/shared/functions.lua:180]: I think we can leave that up to the caller. Surely there are usecases where we want to await for an action with the expectation that it may fail. I don't think we should throw an error in that case. However, if an errorMessage is provided, we should throw an actual error.

[core]/es_extended/shared/functions.lua
```diff
@@ -170,4 +170,52 @@ function ESX.IsFunctionReference(val)
...
+
+    if not isFunctionReference then
+        error('Condition Function should be a function reference.')
+    end
+
+    local startTime = GetGameTimer()
+    local prefix = ('[%s] -> '):format(GetInvokingResource())
+
+    while GetGameTimer() - startTime < timeout do
+        local success, result = pcall(conditionFunc)
```
Comment (Kenshiin13) [inline [core]/es_extended/shared/functions.lua:196]: I don't think we should pcall. If the function throws an error, let it propagate. Especially since we call this every frame.

[core]/es_extended/shared/functions.lua
```diff
@@ -170,4 +170,52 @@ function ESX.IsFunctionReference(val)
...
+        end
+
+        Wait(0)
+    end
+
+    if errorMessage then
+        ESX.AssertType(errorMessage, 'string', 'errorMessage should be a string.')
+
+        local formattedErrorMessage = ('%s %s'):format(prefix, errorMessage)
+        print(formattedErrorMessage)
```
Comment (Kenshiin13) [inline [core]/es_extended/shared/functions.lua:217]: We should throw an actual error here.

[core]/es_extended/shared/functions.lua
```diff
@@ -170,4 +170,52 @@ function ESX.IsFunctionReference(val)
...
+        if result then
+            local elapsedTime = GetGameTimer() - startTime
+            return true, elapsedTime
+        end
+
+        Wait(0)
+    end
+
+    if errorMessage then
+        ESX.AssertType(errorMessage, 'string', 'errorMessage should be a string.')
```
Comment (Kenshiin13) [inline [core]/es_extended/shared/functions.lua:214]: Validate this before running the actual loop (in line 191 for example).

Kenshiin13 (changes requested):

Kenshiin13:
I will do some testing after the latest requested changes have been implemented and merge this. Thank you!

Kenshiin13 (approved):
Thanks!

Mycroft-Studios (approved):

PR #1542, opened 2024-12-15
fix(es_extended/client/modules/actions) EnterVehicle wrong variable name assign

Zykem:
This PR fixes the EnterAborted event in actions.lua. 
Previously, EnterVehicle was incorrectly setting `self.isEnteringVehicle` 
instead of `self.enteringVehicle`, causing an infinite `ESX.PlayerData.vehicle` 
state if the player aborted entering the vehicle.


Kenshiin13 (approved):
LGTM, thanks!

PR #1544, opened 2024-12-15
feat(es_extended/shared): add ox inventory detection back

Mycroft-Studios:
This PR brings ox inventory detection back, because it seems users are unable to read a very simple message


Kenshiin13 (approved):
Thank you.

PR #1547, opened 2024-12-15
fix(es_extended/client/functions): Fixed notification event

LivedBackwardss:
Not sure why this was changed to `ESX.SecureNetEvent` but other resources that call this event themselves doesn't display notifications.

### DIFF
--- a/[core]/es_extended/client/functions.lua
+++ b/[core]/es_extended/client/functions.lua
@@ -1525,11 +1525,11 @@ function ESX.ShowInventory()
     end)
 end
 
-ESX.SecureNetEvent('esx:showNotification', ESX.ShowNotification)
+RegisterNetEvent('esx:showNotification', ESX.ShowNotification)
 
-ESX.SecureNetEvent('esx:showAdvancedNotification', ESX.ShowAdvancedNotification)
+RegisterNetEvent('esx:showAdvancedNotification', ESX.ShowAdvancedNotification)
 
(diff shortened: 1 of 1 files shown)


Mycroft-Studios:
These events are not intended to be called by the client.
for the client, use the ESX functions.
these are purely for Server->Client

LivedBackwardss:
> These events are not intended to be called by the client. for the client, use the ESX functions. these are purely for Server->Client

So, we have to rely on resource developers to update they scripts that are escrowed? Don't get why it has to be like that.

Mycroft-Studios:
> > These events are not intended to be called by the client. for the client, use the ESX functions. these are purely for Server->Client
> 
> So, we have to rely on resource developers to update they scripts that are escrowed? Don't get why it has to be like that.

Its not on us that developers cannot code correctly 🤷🏼

LivedBackwardss:
> > > These events are not intended to be called by the client. for the client, use the ESX functions. these are purely for Server->Client
> > 
> > 
> > So, we have to rely on resource developers to update they scripts that are escrowed? Don't get why it has to be like that.
> 
> Its not on us that developers cannot code correctly 🤷🏼

So, we have to suffer because they refuse to use `ESX.ShowNotification` and the event. Now, we’re forced to wait for them to update every escrowed resource to include that. I don’t understand why compatibility can’t be implemented.

Mycroft-Studios:
>  I don’t understand why compatibility can’t be implemented.

its not compatibility if it was **Never** supposed to be used in that way at all.
if "developers" cannot read documentation, it is not our fault.

PR #1550, opened 2024-12-15
chore(es_extended/server): remove actions.lua

Mycroft-Studios:
This entire file is filled with bad pratices, and is completely useless


Arctos2win:
Where did the loop go ? 
bcs if these events doesnt get triggered then esx_hud will completly break 
also will pretty much break backwards compatibility if they dont get triggered

Mycroft-Studios:
your thinking of the client one,
this is the server one, that quite literally, does nothing

Arctos2win (approved):

Arctos2win:
my bad

PR #1554, opened 2024-12-19
refactor(es_extended/client/imports/point.lua) and (es_extended/client/modules/points.lua): Performace Improvement

feelfreetofee:
**Inside** thread is now handled in the same resource as long more than one point is nearby.

### DIFF
--- a/[core]/es_extended/client/imports/point.lua
+++ b/[core]/es_extended/client/imports/point.lua
@@ -1,30 +1,39 @@
-Point = ESX.Class()
+local Point = ESX.Class()
+
+local nearby, loop = {}
 
 function Point:constructor(properties)
 	self.coords = properties.coords
-	self.hidden = properties.hidden or false
-	self.inside = properties.inside or function() end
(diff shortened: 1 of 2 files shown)


[core]/es_extended/client/imports/point.lua
```diff
@@ -1,44 +1,53 @@
...
-	self.hidden = hidden
-	ESX.HidePointInternal(self.handle, hidden)
-end
-
+local Point = ESX.Class()
+
+local nearby, loop = {}
+
+function Point:constructor(properties)
+	self.coords = properties.coords
```
Comment (Arctos2win) [inline [core]/es_extended/client/imports/point.lua:6]: Maybe do it like this: 

self = {
coords = , etc...
}

Would like to know someone else opinion abt it but i think it would look cleaner

Comment (Mycroft-Studios) [reply [core]/es_extended/client/imports/point.lua:6]: that would defeat the purpose of the class

Mycroft-Studios (approved):

PR #1557, opened 2024-12-21
feat(es_extended) Dynamic Modules

Zykem:
This PR introduces a dynamic module system to the ESX Framework. The key features and purposes of this system are as follows:

- **Dynamic Module Loading:** Allows loading of modules dynamically from specified paths and resources.
- **Client and Server Support:** Supports both client side and server side module loading with valid handling to prevent cross-loading.
- **Enhanced Flexibility:** Enables developers to load custom modules from different resources, enhancing the flexibility and modularity of the ESX Framework.

# Motivation

The dynamic module system aims to improve the modularity and flexibility of the ESX Framework, allowing developers to extend and customize the framework more efficiently. This system is particularly useful for code that does not need to run within the ESX Core but inside the resource that requires it. For example:

- **Blips:** es_extended does not need to store blips created by other resources.
- **Peds:** Custom peds can be managed by specific resources.
- **KVP database wrapper:** Resource-dependent database wrappers can be loaded dynamically.

**Usage isntructions**

- Define your modules in the specified path and ensure they are included in fxmanifest.lua file using the files directive
- Use the ESX.LoadModule function to load modules dynamically by specifying the module name, optional path and optional resource name.
^ You can either return a table in the module and use it like i provided down below with the "db" module, or you can use it like in the blip module, just creating glboal functions.

**Example**

Although you can find an example in client/dynamicModules/blip.lua, i will provide another one:

_Expandable code, just showing what it could be used for_

```lua
local module = {}

module.saveToDb = function(key, value)
    SetResourceKvp(key, value) 
end

module.getFromDb = function(key)
    return GetResourceKvpString(key)
end

return module
```

```lua
-- Loading the module assuming the module is inside es_extended/server/dynamicModules/kvp.lua
local db = ESX.LoadModule('kvp')
db.saveToDb('exampleKey, 'exampleValue')

local exampleValue = db.getFromDb('exampleKey')
```

**Loading a module that is in a different resource**
```lua
-- this module would have to be in esx_testing/modules/testModule.lua
ESX.LoadModule('testModule', 'modules', 'esx_testing')
```
 
```

### DIFF
--- a/[core]/es_extended/client/dynamicModules/blip.lua
+++ b/[core]/es_extended/client/dynamicModules/blip.lua
@@ -1,0 +1,65 @@
+-- inspired by: https://github.com/esx-framework/esx_core/pull/1490
+
+local blips = {}
+
+---@class ESXBlip
+---@field coords vector3 Blip coordinates
+---@field sprite number Blip sprite. See https://docs.fivem.net/docs/game-references/blips/#blips
+---@field label string Blip display name/label
+---@field color? number Blip color. See https://docs.fivem.net/docs/game-references/blips/#blip-colors
+---@field scale? number Blip scale. Defaults to 1.0
+---@field display? number Blip display type. Defaults to 4
+---@field shortRange? boolean Is Blip short range. Defaults to true
+
+--- Used to create static blips on the map. Blips are removed when the resource that created them is stopped.
+---@param blipData ESXBlip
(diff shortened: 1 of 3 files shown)


Mycroft-Studios:
Please check dev branch before PRing, this already exists there

PR #1558, opened 2024-12-23
fix(es_extended/): fix race conditions & ped change not propogating

Kenshiin13:
### Summary
- Fixed an issue where relying on `ESX.PlayerLoaded` to be `true` in importing resources did not garuantee that the ped has fully spawned.
- Fixed an issue with ped changes not propogating to importing resources due to execution order.
- Fixed death module relying on a possibly invalid ped which would trigger the death logic.

---

### PR Checklist

### DIFF
--- a/[core]/es_extended/client/modules/actions.lua
+++ b/[core]/es_extended/client/modules/actions.lua
@@ -46,7 +46,6 @@ function Actions:TrackPed()
 
     if playerPed ~= newPed then
         ESX.SetPlayerData("ped", newPed)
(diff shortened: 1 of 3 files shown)


N0tNvll (approved):
Fixes to: 
- Prevent to change the loaded status to true if the ped isnt loaded.
- Checks to detect if the ped already exist before checking the status.

All of this changes are mainly made to prevent possible issues in different situations like when the player load/reload or status of the ped

PR #1569, opened 2025-01-02
1.12 to 1.12.0 -> lib.checkDependency

userMacieG:


Mycroft-Studios:
why?

userMacieG:
> why?

https://semver.org/ 👍🏽

userMacieG:
before ox inv was having problem detecting current version but now when you actually used semver is fixed

PR #1583, opened 2025-01-06
feat(es_extended/client): add vehicle type to ESX.Game.GetVehicleProp…

someone005:
### Description
This PR adds vehicle type to ESX.Game.GetVehicleProperties return table.

---
### Motivation
Developers won't have to manually add vehicle type to properties table.

---

### PR Checklist
- [ x ]  My commit messages and PR title follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard.
- [ x ]  My changes have been tested locally and function as expected.
- [ x ]  My PR does not introduce any breaking changes.
- [ x ]  I have provided a clear explanation of what my PR does, including the reasoning behind the changes and any relevant context.


Mycroft-Studios:
why would you need type?

someone005:
> why would you need type?

For example, developer saves vehicle properties in database and next wants to create vehicle using CreateVehicleServerSetter. He can take type from properties without manually putting them in properties table. I don't think it important thing but ig that some developers would like it.

PR #1591, opened 2025-01-09
fix(es_extended/server/functions): improve `Core.SavePlayers` & `Core.SavePlayer`

Kenshiin13:
### Description
Improves the performance of the `Core.SavePlayers` & `Core.SavePlayer` functions.

---
### Motivation
Larger servers experience regular server thread hitches when ESX saves all of their players. This PR aims to improve that.

---

- Only save `lastPlayTime` player metadata on player leave.
- Set player metadata directly when saving players. This avoids unnecessary type checks, as well as client events.
- Get player coords directly when saving position. Avoids redundant native calls to get a players ped.
- Added optional new `Core.SavePlayers` function while keeping the old one enabled by default through a new config option. The new one avoids any native calls entirely by sacraficing the players position, health and armour data. This can be saved on player leave, though it may result in data loss when the client or server crashes.
---

### PR Checklist

### DIFF
--- a/[core]/es_extended/fxmanifest.lua
+++ b/[core]/es_extended/fxmanifest.lua
@@ -12,6 +12,7 @@ shared_scripts {
 	'shared/config/main.lua',
     'shared/config/weapons.lua',
     'shared/config/adjustments.lua',
+    'shared/config/compat.lua',
 
     'shared/main.lua',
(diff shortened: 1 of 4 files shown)


Kenshiin13:
Stale PR

PR #1595, opened 2025-01-14
Add ESX.Switch function

Zykem:
### Description

---

This PR adds a new ESX.Switch function. It enhances the existing logic by enabling flexible argument passing to case functions and improving error handling. The function is similar to the `switch` statement found in other programming languages like JavaScript, offering an elegant way to handle multiple conditions. It allows handling various cases based on a given value.

### Motivation

---

Motivation of this PR is to simplify and clean up decision-making logic. Instead of constnatly using if-elseif blocks, it offers a more readable and maintainable way of handling cases.

---

It was implemented to receive a value, a table of possible cases, an optional default function and extra arguments that are passed to the default function (if exists).

Each case is a key-value pair in caseTable, where the key corresponds to a value and the value is a function.

It alsoo allows passing additional parameters to the matching case function by using lua's `...` syntax.

Error handling is already included using core functions like `ESX.IsFunctionReference`, `ESX.AssertType` or lua's `assert`

### Usage Example

---
As an example, i took this piece of code you could use for a player manager resource (similiar to EasyAdmin) where you can ban/kick players. You could use one net-event for all this and just store the actual event functions in the caseTable

```lua
local playerHandlers = {
    ban = function(playerId, reason)
        -- ban logic
    end,
    kick = function(playerId, reason)
        DropPlayer(playerId, reason or 'Unknown reason')
    end
}

RegisterNetEvent('playerActionEvent', function(action, ...)
     local source = source
    ESX.Switch(action, playerHandlers, function()
         -- default function was called, did not find a match!
    end, source, ...)
end)
```

### PR Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -2,8 +2,8 @@
 name: Bug report
 about: Create a report to help us improve
 title: "[Bug] - esx_script - Issue"
-labels: bug, enhancement
-assignees: Kenshiin13, Arctos2win
+labels: bug
+assignees: Arctos2win, Kenshiin13
 
 ---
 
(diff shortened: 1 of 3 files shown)


Arctos2win:
Looks like a very unoptimized way to do an else if statement.

PR #1599, opened 2025-01-19
refactor(es_extended/client/modules/scaleform): Make scaleforms show duration framerate independent

fnbar0:
### Description

- This PR updates the scaleform display logic to ensure that the duration of messages is no longer dependent on the frame rate. The new implementation uses ```GetGameTimer()``` to calculate when the scaleform should stop showing.
---
### Motivation

- The previous implementation of the scaleform showing duration relied on frame-dependent timing, which caused inconsistent behavior on different machines or just different FPS. A 5 second scaleform would hide after ~5 seconds on 100FPS but after ~10 seconds on 50FPS. This change ensures that the showing duration is accurate and not dependent of client framerate.
---

- Replaced the frame-dependent loop ```(sec = sec - 0.01)``` with a time-based calculation using ```GetGameTimer()```.
- The loop now checks the elapsed time in milliseconds instead of decrementing a counter every frame.

---

### PR Checklist

### DIFF
--- a/[core]/es_extended/client/modules/scaleform.lua
+++ b/[core]/es_extended/client/modules/scaleform.lua
@@ -3,11 +3,10 @@ ESX.Scaleform.Utils = {}
 
 function ESX.Scaleform.ShowFreemodeMessage(title, msg, sec)
     local scaleform = ESX.Scaleform.Utils.RunMethod("MP_BIG_MESSAGE_FREEMODE", "SHOW_SHARD_WASTED_MP_MESSAGE", false, title, msg)
-
-    while sec > 0 do
+    
+    local startTimer = GetGameTimer()
+    while GetGameTimer() - startTimer < sec * 1000 do
         Wait(0)
-        sec = sec - 0.01
-
         DrawScaleformMovieFullscreen(scaleform, 255, 255, 255, 255, 0)
     end
 
@@ -19,10 +18,9 @@ function ESX.Scaleform.ShowBreakingNews(title, msg, bottom, sec)
     ESX.Scaleform.Utils.RunMethod(scaleform, "SET_SCROLL_TEXT", false, 0, 0, title)
     ESX.Scaleform.Utils.RunMethod(scaleform, "DISPLAY_SCROLL_TEXT", false, 0, 0)
 
-    while sec > 0 do
+    local startTimer = GetGameTimer()
+    while GetGameTimer() - startTimer < sec * 1000 do
         Wait(0)
-        sec = sec - 0.01
-
         DrawScaleformMovieFullscreen(scaleform, 255, 255, 255, 255, 0)
     end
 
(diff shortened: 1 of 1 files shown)


[core]/es_extended/client/modules/scaleform.lua
```diff
@@ -3,11 +3,10 @@ ESX.Scaleform.Utils = {}
 
 function ESX.Scaleform.ShowFreemodeMessage(title, msg, sec)
     local scaleform = ESX.Scaleform.Utils.RunMethod("MP_BIG_MESSAGE_FREEMODE", "SHOW_SHARD_WASTED_MP_MESSAGE", false, title, msg)
-
-    while sec > 0 do
+    
+    local startTimer = GetGameTimer()
+    while GetGameTimer() - startTimer < sec * 1000 do
```
Comment (Kenshiin13) [inline [core]/es_extended/client/modules/scaleform.lua:8]: No need to compute that multiplication every iteration. Create a var outside the loop.

[core]/es_extended/client/modules/scaleform.lua
```diff
@@ -19,10 +18,9 @@ function ESX.Scaleform.ShowBreakingNews(title, msg, bottom, sec)
     ESX.Scaleform.Utils.RunMethod(scaleform, "SET_SCROLL_TEXT", false, 0, 0, title)
     ESX.Scaleform.Utils.RunMethod(scaleform, "DISPLAY_SCROLL_TEXT", false, 0, 0)
 
-    while sec > 0 do
+    local startTimer = GetGameTimer()
+    while GetGameTimer() - startTimer < sec * 1000 do
```
Comment (Kenshiin13) [inline [core]/es_extended/client/modules/scaleform.lua:22]: No need to compute that multiplication every iteration. Create a var outside the loop.

[core]/es_extended/client/modules/scaleform.lua
```diff
@@ -32,10 +30,9 @@ end
 function ESX.Scaleform.ShowPopupWarning(title, msg, bottom, sec)
     local scaleform = ESX.Scaleform.Utils.RunMethod("POPUP_WARNING", "SHOW_POPUP_WARNING", false, 500.0, title, msg, bottom, true)
 
-    while sec > 0 do
+    local startTimer = GetGameTimer()
+    while GetGameTimer() - startTimer < sec * 1000 do
```
Comment (Kenshiin13) [inline [core]/es_extended/client/modules/scaleform.lua:34]: No need to compute that multiplication every iteration. Create a var outside the loop.

[core]/es_extended/client/modules/scaleform.lua
```diff
@@ -45,10 +42,9 @@ end
 function ESX.Scaleform.ShowTrafficMovie(sec)
     local scaleform = ESX.Scaleform.Utils.RunMethod("TRAFFIC_CAM", "PLAY_CAM_MOVIE", false)
 
-    while sec > 0 do
+    local startTimer = GetGameTimer()
+    while GetGameTimer() - startTimer < sec * 1000 do
```
Comment (Kenshiin13) [inline [core]/es_extended/client/modules/scaleform.lua:46]: No need to compute that multiplication every iteration. Create a var outside the loop.

Kenshiin13 (changes requested):

Kenshiin13 (approved):
LGTM, thanks!

PR #1601, opened 2025-01-22
feat(es_extended): add static player methods

Kenshiin13:
### Description  
Introducing **static ESX Player methods** to streamline player handling. Instead of dealing with a bloated `xPlayer` object that can impact performance as player counts grow, this implementation provides a lightweight alternative by dynamically calling only the required functionality through exports.

---

### Motivation  
The current `xPlayer` object is overly bloated, storing unnecessary data that scales poorly with larger player bases. This approach addresses these performance concerns while maintaining a syntax familiar to ESX developers.

---

### Implementation Details  
The `ESX.Player` object uses a metamethod to return a lightweight table containing only the player's server ID when called. If the requested player is offline, it simply returns `nil`. For any subsequent method calls, the stored server ID is used to trigger the appropriate export, ensuring efficient and dynamic behavior.

---

### Usage Example
```lua
local src = 1
local xPlayer = ESX.Player(src)
if (not xPlayer) then
    print("player not found")
    return
end

print(xPlayer.getIdentifier())

local identifier = "char1:12754a0705e35cda21ced87c7daf329730135087"
local xTarget = ESX.Player(identifier)
if (not xTarget) then
    print("target not found")
    return
end

print(xTarget.getIdentifier())

```
---

### PR Checklist

### DIFF
--- a/[core]/es_extended/imports.lua
+++ b/[core]/es_extended/imports.lua
@@ -47,7 +47,9 @@ else
         __call = function(_, src)
             if type(src) ~= "number" then
                 src = ESX.GetPlayerIdFromIdentifier(src)
-                if not src then return end
+                if not src then
+                    return
+                end
             elseif not ESX.IsPlayerLoaded(src) then
                 return
             end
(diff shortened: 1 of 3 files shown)


Arctos2win (approved):
Looks good

Arctos2win (approved):
👍

Kenshiin13:
https://github.com/esx-framework/docs-rewrite/commit/e039e024afd4e04600ddc68536c273da21bcb15c

PR #1612, opened 2025-02-13
feat(es_extended/client/imports/point.lua): Merge module into imports

feelfreetofee:
### Description
The module is now exclusively imported.

### Motivation
The development team has chosen to restrict the module's scope to the importing resource

I have implemented two separate threads to manage point detection efficiently:

Main Thread:
- Runs every 500ms.
- Starts when at least one point is registered and stops when the last point is deleted.
- Checks if the player enters or leaves any points.

Nearby Thread:
- Starts when the player gets close to any point for the first time.
- Runs every frame and stops when there are no more nearby points.
- Calls the inside method when the player is within range of a point.

This setup ensures that the main thread runs intermittently to save resources, only activating when necessary.
The nearby thread runs at full speed only when the player is near relevant points, ensuring responsive interactions.

### Usage Example
```lua
local pointObject = ESX.Point:new({
   coords = vector3(0, 0, 70),
   distance = 10
})

function pointObject:enter()
   print('Hello World!')
end

function pointObject:inside(distance)
   DrawMarker(1, self.coords.x, self.coords.y, self.coords.z, 0, 0, 0, 0, 0, 0, 2, 2, 1, 255, 255, 255, 151)
   if distance <= 2 then
      ESX.ShowHelpNotification(('Hello World #%d'):format(self.handle), true)
   end
end

function pointObject:leave()
   print('Bye World!')
end
```

### PR Checklist


[core]/es_extended/client/imports/point.lua
```diff
@@ -1,53 +1,76 @@
...
+                    if not point.hidden and #(coords - point.coords) <= point.distance then
+                        if not nearbyPoints[handle] then
+                            nearbyPoints[handle] = point
+                            if point.enter then
+                                point:enter()
+                            end
+                            if not nearbyThread then
+                                nearbyThread = true
+                                CreateThread(function()
+                                    while nearbyThread do Wait()
```
Comment (Mycroft-Studios) [inline [core]/es_extended/client/imports/point.lua:39]: Can you do Wait(0), mainly for the linter.

Mycroft-Studios (changes requested):

[core]/es_extended/client/imports/point.lua
Comment (feelfreetofee) [reply [core]/es_extended/client/imports/point.lua]: Linter issues fixed, thank you for letting me know. Also, great tool .

Kenshiin13:
As discussed, this will become obsolete with the next update. I'm really sorry for the time that has been dedicated for this!

PR #1614, opened 2025-02-17
fix(es_extended/server/functions): missing attributes for unemployed

t1ger-scripts:
Missing attributes when indexing unemployed job into ESX.Jobs. Database contains name column for jobs table  and for job_grades it also contain a grade name column.

Also fixed loop method for grades when inserting into queries.

### Description
Adds the missing attributes when assigning unemployed to ESX.Jobs if no jobs are found. 
Also fixed loop method for grades when inserting into queries.
---

### Motivation
Because it makes sense
---

N/A
---

### Usage Example
3rd party script that compares grades[<grade>].name wouldn't work if this attribute is missing.
Fixes issues where creators parse job grades indexed with strings.
---

### PR Checklist

### DIFF
--- a/[core]/es_extended/server/modules/createJob.lua
+++ b/[core]/es_extended/server/modules/createJob.lua
@@ -75,7 +75,7 @@ function ESX.CreateJob(name, label, grades)
     for _, grade in pairs(grades) do
         queries[#queries + 1] = {
             query = 'INSERT INTO job_grades (job_name, grade, name, label, salary, skin_male, skin_female) VALUES (?, ?, ?, ?, ?, ?, ?)',
-            values = { name, grade.grade, grade.name, grade.label, grade.salary, json.encode(grade.skin_male) or '{}', json.encode(grade.skin_female) or '{}' }
+            values = { name, grade.grade, grade.name, grade.label, grade.salary, grade.skin_male and json.encode(grade.skin_male) or '{}', grade.skin_female and json.encode(grade.skin_female) or '{}' }
         }
     end
 
--- a/[core]/es_extended/server/modules/createJob.lua
+++ b/[core]/es_extended/server/modules/createJob.lua
@@ -72,10 +72,10 @@ function ESX.CreateJob(name, label, grades)
         { query = 'INSERT INTO jobs (name, label) VALUES (?, ?)', values = { name, label } }
     }
 
-    for _, grade in ipairs(grades) do
+    for _, grade in pairs(grades) do
         queries[#queries + 1] = {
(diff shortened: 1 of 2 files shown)


Kenshiin13 (changes requested):
`xPlayer.job.whitelisted` is still missing

t1ger-scripts:
> `xPlayer.job.whitelisted` is still missing

Corrected, thanks.

t1ger-scripts:
> The queries assume that the parsed grades are numerically indexed. But removing ipairs method, and using pairs method, allows to loop grades that are indexed with strings as well. Just like we are using pairs method inside generatenewJobTable. So added this to the PR as well.

[core]/es_extended/server/modules/createJob.lua
```diff
@@ -72,10 +72,10 @@ function ESX.CreateJob(name, label, grades)
         { query = 'INSERT INTO jobs (name, label) VALUES (?, ?)', values = { name, label } }
     }
 
-    for _, grade in ipairs(grades) do
+    for _, grade in pairs(grades) do
         queries[#queries + 1] = {
             query = 'INSERT INTO job_grades (job_name, grade, name, label, salary, skin_male, skin_female) VALUES (?, ?, ?, ?, ?, ?, ?)',
-            values = { name, grade.grade, grade.name, grade.label, grade.salary, '{}', '{}' }
+            values = { name, grade.grade, grade.name, grade.label, grade.salary, json.encode(grade.skin_male) or '{}', json.encode(grade.skin_female) or '{}' }
```
Comment (Kenshiin13) [inline [core]/es_extended/server/modules/createJob.lua:78]: `json.encode(grade.skin_male) or '{}'` 
`json.encode(grade.skin_female) or '{}'`

This won't work as json encoding a nil value results in the string "null", which is not a falsy value.

Kenshiin13 (changes requested):

[core]/es_extended/server/modules/createJob.lua
Comment (Kenshiin13) [reply [core]/es_extended/server/modules/createJob.lua:78]:
```lua
grade.skin_male and json.encode(grade.skin_male) or '{}'
grade.skin_female and json.encode(grade.skin_female) or '{}'
```

Should do the trick.

Comment (t1ger-scripts) [reply [core]/es_extended/server/modules/createJob.lua]: > ```lua
> grade.skin_male and json.encode(grade.skin_male) or '{}'
> ```

Fixed

t1ger-scripts:
> ```lua
> grade.skin_male and json.encode(grade.skin_male) or '{}'
> grade.skin_female and json.encode(grade.skin_female) or '{}'
> ```
> 
> Should do the trick.

Fixed.

PR #1616, opened 2025-02-21
feat(es_extended/client/modules/action): add TrackJumping on SlowLoop…

KrisnaCahya:
addition of a jumping system in slowloop to make things easier when needed, such as event jumping in the previous version

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -2,8 +2,8 @@
 name: Bug report
 about: Create a report to help us improve
 title: "[Bug] - esx_script - Issue"
-labels: bug, enhancement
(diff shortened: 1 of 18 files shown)


KrisnaCahya:
I think this is quite necessary

Kenshiin13:
This would need to run every frame to be reliable. In addition to that, it is too specific for a framework-wide cache for my liking. There's maybe a handful of resources where I could imagine someone would need this, which in turn doesn't justify caching this every frame. Though I appreciate the effort you put into your contribution, I don't think this has a place in the framework.

- Side note:
As mentioned in our [Contribution Guidelines](https://github.com/esx-framework/esx_core/blob/main/CONTRIBUTING.md), we merge into dev branch. Please base your PRs on dev branch in the future.

PR #1626, opened 2025-03-30
fix(config): update minimum age year from 2005 to 2007 for accuracy

PhilCode-creator:
### Description
Just a update to match the right year because its 2025 and you can be 18 years old if you are born in 2007

---
### Motivation
I just opend the config and it bothered me

---

Just a Config value update
---

### Usage Example
None
---

### PR Checklist
- [ ]  My commit messages and PR title follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard.

### DIFF
--- a/[core]/esx_identity/config.lua
+++ b/[core]/esx_identity/config.lua
@@ -14,7 +14,7 @@ Config.MaxNameLength = 20 -- Max Name Length.
 Config.MinHeight = 120 -- 120 cm lowest height
 Config.MaxHeight = 220 -- 220 cm max height.
 Config.LowestYear = 1900 -- 112 years old is the oldest you can be.
-Config.HighestYear = 2005 -- 18 years old is the youngest you can be.
+Config.HighestYear = 2007 -- 18 years old is the youngest you can be.
 
 Config.FullCharDelete = true -- Delete all reference to character.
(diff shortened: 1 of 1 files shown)


YOMAN1792:
Hello, can you change the comment that tells how old you’ll be today, please?

YOMAN1792:
> > Hello, can you change the comment that tells how old you’ll be today, please?
> 
> what exactly do you mean with that? I mean thats the old one or do you mean with the pr

Line 16 of the file in the comment if you can update the age do it can be accurate please.

tomiichx:
You could instead do the calculation automatically by simply specifying the min. age required..? Do you want to keep coming back every year to PR such changes?

Zykem (changes requested):
Hey, thanks for the PR! Can you update the script to automatically set the minimum age to 18, instead of just changing the config value? That way, it will be handled dynamically. We dont want to continue working with the same bad code from the previous team members, so we prefer to upgrade ESX properly.

Eventually, you could create a new config property like:

`Config.MinimumAge = 18`

Then, when registering the identity, you can check the users age to make sure it matches the minimum age set in the config. But we wont approve changes that continue to use this “bad” code.

Zykem:
Hi, we implemented a different solution for this problem. Thank you for the effort you put in this PR.

[New PR](https://github.com/esx-framework/esx_core/pull/1632)

PR #1630, opened 2025-04-23
Refactor name validation: remove numeric check and simplify character validation

YOMAN1792:
### Description
This PR refactors the name validation logic by removing the redundant numeric check and simplifying the character validation to handle various character sets like Latin, Greek, Cyrillic, Hebrew, Arabic, and CJK. This change improves the validation process and ensures that only valid characters are accepted, making it more flexible for different language users within the community.

---
### Motivation
The previous validation logic had a redundant numeric check, which was unnecessary because it was already covered by the new character validation logic. This update simplifies the code, improving code readability and making the validation process more flexible. It better supports a wider range of languages and scripts, which is an important step toward making the framework more inclusive for a global community.

---

- The 'checkValidCharacter()' function is now responsible for validating characters based on their Unicode code ranges, including Latin, Greek, Cyrillic, Hebrew, Arabic, and CJK characters.

- The numeric check has been removed as it was redundant.

- The 'checkNameFormat()' function now solely depends on 'checkValidCharacter()' to validate names.

- This change is also a step toward ensuring that the ESX framework continues to grow in a way that is inclusive and beneficial for developers around the world.
---

### Usage Example
The following function validates a name:
```lua
local function checkNameFormat(name)
    if checkValidCharacter(name) then
        local stringLength = string.len(name)
        return stringLength > 0 and stringLength < Config.MaxNameLength
    end

    return false
end
```
---

### PR Checklist
- [X]  My changes have been tested locally and function as expected.
- [X]  I have provided a clear explanation of what my PR does, including the reasoning behind the changes and any relevant context.

### DIFF
--- a/[core]/es_extended/shared/functions.lua
+++ b/[core]/es_extended/shared/functions.lua
@@ -210,3 +210,72 @@ function ESX.Await(conditionFunc, errorMessage, timeoutMs)
 
     return false
 end
+
+---@param str string
+---@return boolean
+function ESX.IsValidLocaleString(str)
+    local locale = string.lower(Config.Locale)
+
+    local defaultRanges ={
+        {0x0041, 0x005A}, -- Basic Latin uppercase
+        {0x0061, 0x007A}, -- Basic Latin lowercase
+        {0x0020, 0x0020}, -- Space
+        {0x002D, 0x002D}, -- Dash
+        {0x00C0, 0x02AF}  -- Latin Extended
+    }
+
+    local localeRanges = {
+        ["el"] = { {0x0370, 0x03FF} }, -- Greek
+        ["sr"] ={ {0x0400, 0x04FF} }, -- Cyrillic
+        ["he"] ={ {0x05D0, 0x05EA} }, -- Hebrew letters
+        ["ar"] = {
+            {0x0620, 0x063F}, -- Arabic
+            {0x0641, 0x064A},
+            {0x066E, 0x066F},
+            {0x0671, 0x06D3},
+            {0x06D5, 0x06D5},
+            {0x0750, 0x077F},
+            {0x08A0, 0x08BD}
+        },
+        ["zh-cn"] ={ {0x4E00, 0x9FFF} } -- CJK
+    }
+
+    local validRanges = {}
+
+    for i = 1, #defaultRanges do
+        validRanges[#validRanges + 1] = defaultRanges[i]
+    end
+
+    if localeRanges[locale] then
+        for i = 1, #localeRanges[locale] do
+            validRanges[#validRanges + 1] = localeRanges[locale][i]
+        end
+    end
+
+    if Config.ValidCharacterSets then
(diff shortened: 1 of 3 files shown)


Zykem (changes requested):
Do early returns.

Zykem (changes requested):
Move the function `checkValidCharacter` to an ESX core function named `ESX.IsValidLocaleString`.

Do not check all character ranges globally. Instead, determine the character range based on the user's selected locale from the config.

For example, if the locale is set to `zh-cn`, validate only Chinese characters. Do not check for Russian, Arabic, or other unrelated character sets.

However, the following character ranges must always be allowed regardless of the selected locale:
- Basic Latin uppercase (A–Z)
- Basic Latin lowercase (a–z)
- Space and dash (` `, `-`)
- Latin Extended

[core]/es_extended/shared/functions.lua
```diff
@@ -210,3 +210,72 @@ function ESX.Await(conditionFunc, errorMessage, timeoutMs)
...
+                    validRanges[#validRanges + 1] = localeRanges[charset][i]
+                end
+            end
+        end
+    end
+
+    for _, code in utf8.codes(str) do
+        local isValid = false
+
+        for _, range in ipairs(validRanges) do
```
Comment (Zykem) [inline [core]/es_extended/shared/functions.lua:268]: Replace ipairs with numeric loop for better performance in validation check
define `range` as validRanges[i] instead of accessing validRanges[i] twice (in the numeric loop.)

[core]/es_extended/shared/functions.lua
```diff
@@ -210,3 +210,72 @@ function ESX.Await(conditionFunc, errorMessage, timeoutMs)
 
     return false
 end
+
+---@param str string
+---@return boolean
+function ESX.IsValidLocaleString(str)
```
Comment (Zykem) [inline [core]/es_extended/shared/functions.lua:216]: Validate type of `str` with ESX.ValidateType

Zykem (changes requested):
Check single comments.

[core]/es_extended/shared/functions.lua
```diff
@@ -210,3 +210,77 @@ function ESX.Await(conditionFunc, errorMessage, timeoutMs)
...
+            {0x06D5, 0x06D5},
+            {0x0750, 0x077F},
+            {0x08A0, 0x08BD}
+        },
+        ["zh-cn"] ={ {0x4E00, 0x9FFF} } -- CJK
+    }
+
+    local validRanges = {}
+
+    for i = 1, #defaultRanges do
```
Comment (Zykem) [inline [core]/es_extended/shared/functions.lua:249]: Define validRanges with `local validRanges = { table.unpack(defaultRanges) }`

Zykem (approved):
LGTM

Zykem (approved):
LGTM, Thanks!

Zykem:
Documentation update regarding your Pull Request.
https://github.com/esx-framework/docs-rewrite/commit/f0dd53e79d9f8bc62db4cb36b0e32bb012a5c08c

PR #1650, opened 2025-06-07
feat(es_extended) Add Modified Job Tracker In GlobalState

Meyndflay:
- Jobs from the SQL database shall now be automatically loaded upon resource start into GlobalState.
- Each job has its own statebag that returns a table including: jobName, jobLabel, jobRoster, jobCount.

### Description

This PR adds a modified job tracker to GlobalState.
---
### Motivation

I drafted this during es_extended version 1.10.5. I saw that the original code already had a GlobalState job tracker and I thought that it was a good start, but I figured that it could be improved to add more functionality. At the time I needed to easily track a job's players without having to make custom functions in every script that I had to iterate through ESX.GetExtendedPlayers() and track their job status, count, etc.
---

I decided to improve upon the current idea of tracking jobs through the use of the GlobalState by first loading every job from the SQL database onto our GlobalState. After that, I made use of existing ESX event handlers to track and update the job's statebag to return not only the count of the job's player count, but also the server IDs of the players, so we could easily get the xPlayer of each member easily.
---

### Usage Example

```lua
-- Sample Usage:
local jobName = "mechanic"

GlobalState[("job:%s"):format(jobName)]                     -- Each Job Has A Global State Bag Formatted As job:jobName

-- Return Value
local xPlayer = ESX.GetPlayerFromId(1)                      -- Let's Get The Framework Table Of Player 1
local job = xPlayer.job

local retVal = GlobalState[("job:%s"):format(job.name)]   

-- retVal Shall Return A Table With These Key-Value Pairs
retVal = {                                                  -- Returned Table
    jobName = "mechanic",                                   -- Job Name
    jobLabel = "Mechanic",                                  -- Job Label
    jobRoster = { 1, 3, 4, 5 },                             -- Table Of Mechanic Player Ids
    jobCount = 4,                                           -- Returns The Number Of The Job Roster
}
```
---

### PR Checklist

### DIFF
--- a/[core]/es_extended/server/main.lua
+++ b/[core]/es_extended/server/main.lua
@@ -1,6 +1,7 @@
 SetMapName("San Andreas")
 SetGameType("ESX Legacy")
 
+local oxmysql = exports.oxmysql
 local oneSyncState = GetConvar("onesync", "off")
 local newPlayer = "INSERT INTO `users` SET `accounts` = ?, `identifier` = ?, `group` = ?"
 local loadPlayer = "SELECT `accounts`, `job`, `job_grade`, `group`, `position`, `inventory`, `skin`, `loadout`, `metadata`"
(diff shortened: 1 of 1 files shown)


Meyndflay:
Ah I see. I didn't know there was a stability issue regarding GlobalState usage on high player count servers.

PR #1651, opened 2025-06-09
feat(es_extended): Optimize Tracking Of Ped Coords

Meyndflay:
### Description

Optimizes the tracking of ped coords from ESX.PlayerData.coords
---
### Motivation

Updating the ESX.PlayerData.coords via a while loop adds CPU msec overhead, that's why we aim to optimize its tracking.
---

Since the main function of ESX.PlayerData.coords is to return the coords of the player's ped, we can use Lua's metatables to alter the function of ESX.PlayerData.coords.
---

### Usage Example

```lua
-- Old version
print(ESX.PlayerData.coords) -- This returns the coords of the player's ped as the .coords key is updated every interval. (Not efficient)

-- New version
print(ESX.PlayerData.coords) -- Still returns the coords of the player's ped but makes use of metamethods to get the ped's coords. (More efficient and only called as needed)
```
---

### PR Checklist

### DIFF
--- a/[core]/es_extended/imports.lua
+++ b/[core]/es_extended/imports.lua
@@ -4,6 +4,10 @@ ESX.currentResourceName = GetCurrentResourceName()
 OnPlayerData = function (key, val, last) end
 
 local function TrackPedCoordsOnce() 
+    if not ESX or not ESX.PlayerData then
+        return
+    end
+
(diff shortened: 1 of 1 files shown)


Zykem (approved):
LGTM, Thanks

PR #1661, opened 2025-07-06
fix(esx_multicharacter): Await nuiReady before setting up characters

Zykem:
### Description
This PR fixes an issue where players would sometimes be stuck with no UI during character selection, especially on slower systems. The root cause is the ToggleMulticharacter event being sent before the NUI is fully loaded. To fix this, we now send a nuiReady event from the React app as soon as it mounts, letting the Lua side know it's safe to proceed.

---
### Motivation
On slower computers, the frontend may not be ready by the time the ToggleMulticharacter event is triggered, causing the UI to either not show up at all or behave incorrectly. This is a frustrating experience during character selection. By waiting for an explicit nuiReady signal from the NUI, we can avoid these race conditions and ensure the UI only shows when it’s actually ready.

---

- Added a useEffect in App.tsx that dispatches a nuiReady event when the React app mounts.
- Lua should now listen for nuiReady before sending any data or showing the UI.
- This ensures that character selection only starts once the frontend is fully initialized.
---

### Usage Example

---

### PR Checklist
- [1]  My commit messages and PR title follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard.
- [1]  My changes have been tested locally and function as expected.
- [1]  My PR does not introduce any breaking changes.
- [1]  I have provided a clear explanation of what my PR does, including the reasoning behind the changes and any relevant context.


Kenshiin13 (approved):
LGTM, thanks!

PR #1662, opened 2025-07-06
fix(esx_multicharacter): cleanup stale player obj

Kenshiin13:
### Description
Clean up stale player entries even for players who crashed during multichar select.

---
### Motivation
We already introduced a cleanup mechanism for players who already chose a character, but no such mechancism existed for players who crashed during char select.

---

Save source to ESX.Players table so we can later test if that source is still an active player. Otherwise we can initiate cleanup of the stale player obj.
---

### PR Checklist


Zykem (approved):
LGTM, Thanks!

Issue #1646, opened 2025-05-26
[Bug] - Vehicles Not Spawning in Correct Routing Bucket


cesarmr-github:
**Describe the bug**
When using the /car command in any dimension other than 0, the vehicle incorrectly spawns in dimension 0 instead of the player's current dimension. This causes unexpected behavior, especially on servers using multiple dimensions or instances.

**To Reproduce**

Log into the server.

Change to a different dimension (e.g., dimension 1 or higher).

Use the command /car [model].

The vehicle will spawn in dimension 0 instead of the current one.

**Expected behavior**
The vehicle should spawn in the same dimension as the player who issued the command.

**Screenshots**

https://streamable.com/pbd13m

**Debug Info (please complete the following information):**

OS: Windows

FiveM Artifact: 13389

ESX Version: 1.12.4

**Additional context**
This issue can cause confusion among players and result in "ghost vehicles" in other dimensions. It's recommended to review the routing bucket or instance handling when spawning the vehicle to ensure it matches the player's current dimension.

Mycroft-Studios:
quick question, does this happen on both base cars *and* custom cars? or just one?

cjosephdc:
Im facing same issue. any fix?

Kenshiin13:
Fixed in https://github.com/esx-framework/esx_core/pull/1666

PR #1666, opened 2025-07-06
fix(es_extended/server/modules/onesync): decouple player proximity from veh spawning

Kenshiin13:
### Description
This PR refactors `ESX.OneSync.SpawnVehicle` to decouple vehicle spawning from the presence or proximity of any specific player, and introduces a new optional `routingBucket` parameter to `ESX.OneSync.GetClosestPlayer`.

Previously, we attempted to get the vehicle type from the closest player and then waited for that player to become the network owner (regardless of what dimension they are in). This caused issues when spawning vehicles in other dimensions, specifically when the player was not in dimension 0—blocking vehicle creation indefinitely. Since we spawn vehicles in dimension 0, getting a closest player that is not in that dimension will make it impossible for them to become the net owner of that vehicle and so the spawning will fail.

Now, the logic is much more robust:
- If a `vehicleType` is not provided, the server falls back to the first available player (if any) to determine it. Regardless of dimension or distance.
- If **no players are online** and no `vehicleType` is provided, spawning fails early with a clear error.
- Otherwise, vehicle creation proceeds without awaiting a network owner. If a player is nearby in dimension 0, ownership is handled normally. If not, the vehicle still spawns and the caller can set the vehicle into the correct dimension.
- Vehicle properties will still be applied once a player enters scope.

---

### Motivation
Issue https://github.com/esx-framework/esx_core/issues/1646

This change resolves a long-standing issue where `/car` and similar commands would fail in non-zero dimensions. By removing the requirement for a dimension 0 player and falling back to any player for type detection, vehicle spawning becomes reliable and dimension-agnostic.

Additionally:
- The vehicle spawn logic was far too limiting. To determine a vehicle type, a player doesn't need to be in scope of the vehicle, yet the old logic enforced that constraint without justification.
- The only case where awaiting a network owner makes sense is to apply vehicle properties immediately. Now, if no one is nearby, we skip waiting—properties will be synced automatically as soon as a player enters scope.

---

- Refactored `SpawnVehicle` to no longer rely on a closest player being in a specific dimension.
- Fallback to the first available player only if `vehicleType` is not explicitly provided.
- Added early rejection with a clear error if no players are online and no `vehicleType` is provided.
- Skipped awaiting a net owner if none is nearby, improving responsiveness.
- Added optional `routingBucket` filter to `ESX.OneSync.GetClosestPlayer` to support scoped searches.
- Ensured that even when no one is in scope, vehicle spawning and eventual property syncing still work reliably.

---

### PR Checklist

### DIFF
--- a/[core]/es_extended/server/functions.lua
+++ b/[core]/es_extended/server/functions.lua
@@ -401,6 +401,8 @@ function ESX.GetVehicleType(model, player, cb)
         elseif cb then
             cb(result)
         end
+
+        return result
     end
 
     model = type(model) == "string" and joaat(model) or model
--- a/[core]/es_extended/server/modules/onesync.lua
+++ b/[core]/es_extended/server/modules/onesync.lua
@@ -114,10 +114,11 @@ function ESX.OneSync.SpawnVehicle(vehicleModel, coords, heading, vehicleProperti
 
     CreateThread(function()
         if not vehicleType then
-            local src, xPlayer = next(ESX.Players)
-
-            if xPlayer then
-                vehicleType = ESX.GetVehicleType(vehicleModel, xPlayer.source)
+            for src, xPlayer in pairs(ESX.GetExtendedPlayers()) do
+                if(GetPlayerPing(src) > 0) then
+                    vehicleType = ESX.GetVehicleType(vehicleModel, xPlayer.source)
+                    break
+                end
             end
         end
 
--- a/[core]/es_extended/server/modules/onesync.lua
+++ b/[core]/es_extended/server/modules/onesync.lua
@@ -114,11 +114,9 @@ function ESX.OneSync.SpawnVehicle(vehicleModel, coords, heading, vehicleProperti
 
     CreateThread(function()
         if not vehicleType then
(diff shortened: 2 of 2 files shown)


rodrigofplourenco (commented):
Tested this locally using a custom dev resource to simulate various scenarios.

Everything behaves as expected:

- Vehicle spawning no longer depends on players being in dimension 0
- Fallback to first available player works when no vehicleType is passed
- Spawning in custom routingBuckets works without ownership issues
- Vehicles spawn even if no one is nearby; properties sync correctly once someone enters scope
- Proper early rejection when no players are online and no vehicleType is provided
- Manual override of vehicleType also works as expected (`automobile`, etc.)

Reviewed the commit history and logic changes: all consistent with the PR goal. No regressions or unexpected behavior found so far.

The only thing left to confirm is routingBucket fallback behavior with multiple players. I wasn’t able to test that part alone, so we should get at least one more person to validate it in a multiplayer setup before merging.

Otherwise, this looks solid.

FBFezz (approved):
LGTM

PR #1668, opened 2025-07-11
Update callback.lua

cjosephdc:
### Description
This PR modifies the Callbacks:Execute method to improve error handling when a callback fails.
---
### Motivation
Using error() after pcall() defeats the purpose of safe error handling and can cause disruptive stack traces or unwanted crashes during runtime (e.g., when spawning an invalid vehicle model). This change allows errors to be logged without halting execution.

---

- Replaced error(errorString) with print() to prevent script crashes when a callback throws an error.
- Ensures callback execution failures are logged cleanly without interrupting overall execution flow.
---


Kenshiin13:
I don't see any issues with throwing an error. Setting currentId to nil on failure is a good catch though.

Kenshiin13 (approved):
LGTM, thanks.

PR #1671, opened 2025-07-13
feat(es_extended): refresh/add items during runtime

Kenshiin13:
### Description
Adds support for creating and refreshing items at runtime.

---

### Motivation
Improves development workflows by allowing items to be added or refreshed dynamically.  
While it's not recommended to refresh items in prod, third-party scripts can call `ESX.AddItems` during startup to register required items without relying on a manual `items.sql` import.

---

### Usage

```lua
    ESX.AddItems({
		{
			name = "water", label = "Water"
		},
		{
			name = "pd_gadget", label = "PD-Gadget", weight = 10, rare = true, canRemove = false
		},
		{
			name = "bread", label = "Bread"
		},
	})
```

---

### PR Checklist
- [x] My changes have been tested locally and work as intended.
- [x] This PR does not introduce any breaking changes.
- [x] I’ve included a clear explanation and context for the changes in this PR.

### DIFF
--- a/[core]/es_extended/client/functions.lua
+++ b/[core]/es_extended/client/functions.lua
@@ -85,7 +85,7 @@ end
 function ESX.SetPlayerData(key, val)
     local current = ESX.PlayerData[key]
     ESX.PlayerData[key] = val
-    if key ~= "inventory" and key ~= "loadout" then
+    if key ~= "loadout" then
         if type(val) == "table" or val ~= current then
             TriggerEvent("esx:setPlayerData", key, val, current)
         end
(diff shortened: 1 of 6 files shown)


Kr3mu (changes requested):
Add correct locales when registering refresh items command.
If you want you can also includes those micro optimizations:
1. Use numeric loop instead of ipairs
2. Unpack a table using indexes instead of "unpack" function
Also you can add notifying player how many items there are in the database. How many are loaded.
Edit:
Check if item does exist if it does then skip to continue (check only name type first  and if the item doesn't exist check other types)
Use prepare instead of query in refreshitems
^ zykem
<details> 
  <summary>Spoiler </summary>
   Use snake case in local variables 🔥
</details>

Kr3mu (approved):
lgtm

PR #1679, opened 2025-07-25
feat(functions.lua): add ESX.GetJob() function to retrieve job data b…

MrSne7aky:
Adds a new function `ESX.GetJob(jobName)` that returns the full job data for a given job name from the `ESX.Jobs` table.

If the job name is invalid or not found, the function returns nil and logs a warning.

This provides a simple way to access job metadata by name.

---

### Description
Adds a utility function to retrieve a single job's full data using its name.

---

### Motivation
To provide a cleaner and more consistent way to access job data, especially in cases where only one specific job is needed.

---

### Implementation Details
- Looks up the job in `ESX.Jobs` by key.
- If found, returns the full job table.
- If not found, returns nil and prints a warning to the console.

---

### Usage Example
```lua
local job = ESX.GetJob("police")
if job then
    print(job.name, job.label)
end
```

### PR Checklist

### DIFF
--- a/[core]/es_extended/server/functions.lua
+++ b/[core]/es_extended/server/functions.lua
@@ -561,6 +561,18 @@ function ESX.GetJobs()
     return ESX.Jobs
 end
 
+---@param jobName string
+---@return table|nil
+function ESX.GetJob(jobName)
+    local job = ESX.Jobs[jobName]
+    if job then
+        return job
+    else
+        print(("[^3WARNING^7] Attempting to get invalid job: ^5%s^7"):format(jobName))
+        return nil
+    end
+end
+
 ---@return table
 function ESX.GetItems()
     return ESX.Items


[core]/es_extended/server/functions.lua
```diff
@@ -561,6 +561,18 @@ function ESX.GetJobs()
     return ESX.Jobs
 end
 
+---@param jobName string
+---@return table|nil
```
Comment (Kenshiin13) [inline [core]/es_extended/server/functions.lua:565]: Use
```lua
---@return table?
```

[core]/es_extended/server/functions.lua
```diff
@@ -561,6 +561,18 @@ function ESX.GetJobs()
...
+function ESX.GetJob(jobName)
+    local job = ESX.Jobs[jobName]
+    if job then
+        return job
+    else
+        print(("[^3WARNING^7] Attempting to get invalid job: ^5%s^7"):format(jobName))
```
Comment (Kenshiin13) [inline [core]/es_extended/server/functions.lua:571]: Simply attempt to retun ESX.Jobs[jobName]. This function should be one line. It is expected that this function will be called via user input a lot, so having a warning printed can be quite annoying.

Kenshiin13 (changes requested):

Kenshiin13:
We can also add an optional `key` param to the function, which would be more aligned with your initial PR.

```lua
ESX.GetJob(jobName: string, key: string?)
```
```lua
local pdLabel = ESX.GetJob("police", "label")
```

PR #1694, opened 2025-08-06
Add a configurable position for the esx_notify

alexistb2904:
### Description

Add in the Config.lua a possibility to configure where do we want the notify to go (top-right, top-left, top-middle, bottom-right, bottom-left, bottom-middle, middle-left, middle-right)

---

### Motivation

On my server an interface was on the right so we can't see the notify, now we can put it in 8 different position

---

### Implementation Details

- Add a line into the Config.lua to choose the position that we want
- In the Notify.lua pass the Config.position when i notification is sent to the NUI
- In the script.js add the class that we choose in the Config.position to the #root element because he's the one that control the position of the notify
- In the style.css add all the variation of position and keyframes animation for the #root element

- Fallback in the Notify.lua to "middle-right" when the Config.position is not set, same into the script.js

---

### Usage Example

Same usage as usual

```lua
ESX.ShowNotification('No Waypoint Set.', true, false, 140)
```

---

### PR Checklist

-   [ ] My commit messages and PR title follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard.
-   [ ] My changes have been tested locally and function as expected.
-   [ ] I have provided a clear explanation of what my PR does, including the reasoning behind the changes and any relevant context.

### DIFF
--- a/[core]/esx_notify/Config.lua
+++ b/[core]/esx_notify/Config.lua
@@ -1,3 +1,4 @@
 Config = {}
 
 Config.notificationSoundEnabled = true 
+Config.position = "middle-right" -- top-right, top-left, top-middle, bottom-right, bottom-left, bottom-middle, middle-left, middle-right
--- a/[core]/esx_notify/Notify.lua
+++ b/[core]/esx_notify/Notify.lua
@@ -37,7 +37,8 @@ local function Notify(notificatonType, length, message, title)
         length = length or 5000,
         message = message or "ESX-Notify",
         title = title or "New Notification",
-        notificationSoundEnabled = Config.notificationSoundEnabled
+        position = Config.position or "middle-right",
+        notificationSoundEnabled = Config.notificationSoundEnabled or false
     }))
 end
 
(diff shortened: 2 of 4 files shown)


Zykem (changes requested):
Thanks for the PR!

Could you adjust the implementation slightly? Specifically:

Add a new `position `parameter to the function `Notify`.
(Please keep in mind that we also have `ESX.ShowNotification` in the core.

If position is provided and matches one of the supported positions, use it.

Otherwise, fall back to your new Config.position.

This will make the function more flexible while keeping the config value as a default. Let me know if you have any questions!

Zykem (changes requested):
Please add the parameter in ESX:ShowNotification (es_extended/client/functions.lua) too, and looks good to go.

Zykem (changes requested):
One last change: add `title` param to self.showNotification in the player class and this PR is good to go. (Make sure its in the right order, so title first, then position.

Zykem:
That would be the expected result. Take your time, enjoy your vacation!

PR #1715, opened 2025-08-30
add a new argument on the register command feature

OualiS:
### Description

This PR introduces **per-argument validation** support in `ESX.RegisterCommand`.
It allows developers to define an `isValid` function (and optional error message) directly inside each argument definition, enabling more granular control and safer command usage.

As an example, the `/giveitem` command now enforces that the `count` argument must be a positive number.

---

### Motivation

Previously, command validation in ESX was limited:

* Only basic type casting (`number`, `string`, `player`, etc.) was enforced.
* Additional constraints (like positive-only numbers) had to be handled manually in the callback.

This led to **duplicated logic** across commands and inconsistent error handling.

By adding `isValid` support, validation becomes **declarative**, reusable, and consistent across the framework.

---

### Implementation Details

* Updated the argument parsing logic inside `ESX.RegisterCommand`:

  * After type parsing, the final parsed value (`newArgs[v.name]`) is passed into `v.isValid`.
  * `pcall` is used to protect against errors inside the validator function.
  * A custom error message can be provided via `v.error`, otherwise a generic localized error is used.
  * Validation runs only if no prior error was detected for the argument.

* Backwards compatibility is maintained:

  * Existing commands without `isValid` continue to work.
---

### Usage Example

```lua
ESX.RegisterCommand(
    "giveitem",
    "admin",
    function(xPlayer, args)
        args.playerId.addInventoryItem(args.item, args.count)
    end,
    true,
    {
        help = "Give an item to a player",
        validate = true,
        arguments = {
            { name = "playerId", type = "player" },
            { name = "item", type = "item" },
            {
                name = "count",
                type = "number",
                isValid = function(n) return type(n) == "number" and n > 0 end,
                err = TranslateCap("commanderror_argumentmismatch_positive_number", "count")
            },
        },
    }
)
```

---

### PR Checklist

* [x] My commit messages and PR title follow the [[Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)](https://www.conventionalcommits.org/en/v1.0.0/) standard.
* [x] My changes have been tested locally and function as expected.
* [x] My PR does not introduce any breaking changes.
* [x] I have provided a clear explanation of what my PR does, including the reasoning behind the changes and relevant context.

### DIFF
--- a/[core]/es_extended/server/functions.lua
+++ b/[core]/es_extended/server/functions.lua
@@ -147,11 +147,11 @@ function ESX.RegisterCommand(name, group, cb, allowConsole, suggestion)
                             end
                         end
                         
-                        if v.isValid and not err then
+                        if ESX.IsFunctionReference(v.Validator?.validate) and not err then
                             local candidate = newArgs[v.name]
-                            local ok, res = pcall(v.isValid, candidate)
+                            local ok, res = pcall(v.Validator.validate, candidate)
                             if not ok or res ~= true then
-                                err = v.err or TranslateCap("commanderror_argumentmismatch")
+                                err = v.Validator.err or TranslateCap("commanderror_argumentmismatch")
                             end
                         end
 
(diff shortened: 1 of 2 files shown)


Kenshiin13:
https://github.com/esx-framework/esx-legacy-documenation/commit/cc2c8a396a882983284f650ffb3d1a6a115eea31

PR #1725, opened 2025-09-14
feat(xPlayer): made xPlayer's functions hookable

HR0110:
### Description
This pull request changes the whole xPlayer code and the code around it, it fixes the code with xPlayer, optimizes it a bit and adds a new feature to the whole ESX table, ESX.XPlayerClass which is a metatable value which has .addHook and .removeHook fields that allow to add hooks to specific functions in the xPlayer, usable in all scripts that use es_extended's xPlayer. It allows the developers to change the code of the functions entirely or particulary. To the .addHook field, there is a isJustAddition parameter that sets whether or not the provided hook function is to replace the whole function or just be as addition that can stop the original function execute by returning `false`. This is going to be very helpful to any developer, working with ESX framework.

### Usage Example
```lua
local xPlayer = ESX.Player(source)

xPlayer.addMoney(250) -- There is no output

local hookExample = ESX.XPlayerClass.addHook('addMoney', true, function(xPlayer, amount)
    print(('Player %s received $%s as cash.'):format(GetPlayerName(xPlayer.source), amount))
end)

xPlayer.addMoney(300) -- Output: `Player example received $300 as cash.`

local success = ESX.XPlayerClass.removeHook(hookExample)
print(success) -- And because the hook exists, the output will be `true`

xPlayer.addMoney(350) -- There is no output
```

### PR Checklist

-   [X] My changes have been tested locally and function as expected.
-   [X] My PR does not introduce any breaking changes.
-   [X] I have provided a clear explanation of what my PR does, including the reasoning behind the changes and any relevant context.


FBFezz (approved):
Changes needed

PR #1739, opened 2025-10-20
fix(commands/tp) Fix teleporting in vehicle

Zykem:
### Description

This PR fixes an issue where admins using the /tp command while in a vehicle would lose their vehicle. Now, if a player is in a vehicle, the vehicle is teleported along with the player.

---

### Motivation

Admins using the /tp command while inside a vehicle would lose the vehicle because only the player’s coordinates were updated. This fix ensures that the vehicle moves along with the player, preventing accidental loss.

---

### Implementation Details

- Retrieve the player ped using GetPlayerPed.
- Check if the player is currently in a vehicle using GetVehiclePedIsIn.
- If a vehicle exists, use SetEntityCoords to teleport the vehicle, which automatically teleports the player as well.
- If the player is not in a vehicle, update the player’s coordinates using xPlayer.setCoords.

---

### Usage Example

```lua
/tp [x] [y] [z]
```

---

### PR Checklist

### DIFF
--- a/[core]/es_extended/server/modules/commands.lua
+++ b/[core]/es_extended/server/modules/commands.lua
@@ -2,14 +2,22 @@ ESX.RegisterCommand(
     { "setcoords", "tp" },
     "admin",
     function(xPlayer, args)
-        xPlayer.setCoords({ x = args.x, y = args.y, z = args.z })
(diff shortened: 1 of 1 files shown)


N0tNvll (approved):

FBFezz (approved):
LGTM

PR #1740, opened 2025-10-20
feat(esx_identity/web): New UI

Zuntie:
### Description

Wrote a new UI for the esx_identity.

---

### Motivation

Tasked. Revamps the old design.

---

### Implementation Details

Implemented using Preact, Vite & TailwindCSS. Furthermore, most of the code is written in TypeScript.

---

### Usage Example

Create a new character.

---

### PR Checklist


Zuntie:
Woops. Accidentally duplicated files from the old version. Ill handle it.

Zuntie:
Finished.

N0tNvll:
Needs to fix the button to create the character sometimes it is out of the container

N0tNvll (approved):

Kr3mu (changes requested):
The UI is too small ( especially the text ).
The bug with the height number:
image
The X mark on date is not centered:
image
The Birth date is not opening, tried waterfox and chromium browser (chromium).
Use vh instead of vw as if someone have ultra wide monitor the ui will be too big.
Use GetCurrentResourceName instead of hardcoding esx_identity
What's more your date validation just does year - 18 but what if you put december, you are not 18.

Kr3mu:
After posting I have noticed that look how much free space there is for example in select of gender

Kr3mu:
To make it look better you can add simple opacity animation or other (just make it quick so people don't have to wait) - not really needed

PR #1741, opened 2025-10-21
feat(esx_identity): implement nationalities system

SNAILX808:
### Description
implemented nationalities system

---
### Implementation Details
- Integrated nationality data into player data structure in es_extended
- Nationality stored as string identifier in database for flexibility

---
### Usage Example
```lua

xPlayer.get("nationality")
----------------

Player(source).state.nationality
----------------

xPlayer.set("nationality", "nana")
```

---
### PR Checklist


[core]/esx_identity/server/main.lua
```diff
@@ -114,6 +116,13 @@ local function checkHeightFormat(height)
     return numHeight >= Config.MinHeight and numHeight <= Config.MaxHeight
 end
 
+local function checkNationalityFormat(nationality)
```
Comment (Zuntie) [inline [core]/esx_identity/server/main.lua:119]: This function will reject nationalities like "Myanmar (Burma)" and "Congo (Democratic Republic of the)" because `ESX.IsValidLocaleString` doesn’t allow parentheses ( or ) in the strings.

Comment (Zuntie) [reply [core]/esx_identity/server/main.lua:119]: Nationalities used by the new UI is as following:
https://github.com/esx-framework/esx_core/pull/1740/files#diff-fe0cd305d882c8897708389328b6ccec88310ec275524628ac68a16bb890f410
Should we change those to match the current check? Or, should the check be edited?

Zuntie:
A nationalities configuration file could be in order. Then the UI will just fetch from it, and a check can be made server-sided.
Currently, an arbitrary string could be passed by a request, and it would still go through.

[core]/esx_identity/server/main.lua
Comment (SNAILX808) [reply [core]/esx_identity/server/main.lua:119]: yes, we should change them to match the current check

PR #1749, opened 2025-10-25
moved ESX.Streaming to esx_lib and made new function requestAudioBank

Stane034:


Kr3mu (changes requested):
We would like to mantine a function functioName standard instead of functionName = function across the lib, so if you could change it. Also you do checking for modelHash in requestmodel but not in the requestWeaponAsset. Instead of adding iferrorMessage you can just do {"string", "nil"}. In my opinion and default timeout of 10000 or don't make it optional.

Kr3mu:
nvm make it optional because of backwards compatibility

PR #1752, opened 2025-10-28
Refactor keybind registration to use xLib.addKeybind

SNAILX808:


[core]/esx_lib/imports/addKeybind/client.lua
```diff
@@ -0,0 +1,107 @@
...
+        if not keybinds[data.name] then return end
+        if data.onPressed then data:onPressed() end
+    end)
+
+    RegisterCommand('-' .. data.name, function()
+        if data.disabled or IsPauseMenuActive() then return end
+        data.isPressed = false
+        if not keybinds[data.name] then return end
+        if data.onReleased then data:onReleased() end
+    end)
```
Comment (Zykem) [inline [core]/esx_lib/imports/addKeybind/client.lua:80]: Move the `keybinds[data.name]` condition on top of the press and release command handlers because it takes less (the other if statement has a native call)

Overall it doesnt make sense to set `isPressed` before the `keybinds[data.name]` if statement

Zykem (changes requested):

Zykem (approved):
LGTM, Tested

PR #1766, opened 2025-12-07
Refactor: Improve project structure and refactor registerIdentity

rwixy:
### Description

This pull request refactors esx_identity by modularizing the codebase, improving the structure, and applying the DRY principle.
The esx_identity:registerIdentity callback has been cleaned up and optimized, and related logic has been moved into dedicated files for better maintainability.

### PR Checklist


ASTROWwwW:
Nice cleanup — the modular split makes `esx_identity` much easier to follow. Two things worth a look:

**1. `completeRegistration` now fires `esx_identity:completedRegistration` in the standalone path too.** In the current `main.lua`, `esx_identity:completedRegistration` is triggered *only* in the multichar branch (`xPlayer == nil`). Here `Callback.RegisterIdentity` calls the shared `completeRegistration()` in **both** branches, so a normal non-multichar registration now also emits that event. Anything subscribed to it (esx_multicharacter, third-party resources) will start receiving it in standalone mode — intended, or should `completeRegistration` skip the event in the `xPlayer` branch to keep the previous behavior?

**2. Two existing gaps you could close while you're in this code:**
- `validateRegistrationData` reads `data.firstname` directly, so a client triggering `esx_identity:registerIdentity` with a non-table `data` throws and aborts the callback — a `type(data) ~= 'table'` guard at the top of `Callback.RegisterIdentity` covers it.
- `CheckSexFormat` accepts `"M"/"F"` but `getFormattedIdentity` stores `data.sex` raw, so `"M"` gets persisted and breaks downstream `sex == "m"` checks; lowercasing before save fixes it.

(I have those two in #1787 against `main` — happy to fold them into `validation.lua`/`callback.lua` here instead.)

Minor: `registry.lua` makes `Modules` a global; a local returned/exported from the module would avoid the global and the implicit fxmanifest load-order coupling.

PR #1775, opened 2026-03-30
feat(es_extended): support weight or limit inventory mode

mysbryce:
### Description

Adds configurable inventory handling to `es_extended`, allowing servers to switch between `weight`-based and `limit`-based item restrictions through config.

---

### Motivation

Some servers still rely on the classic per-item `limit` behavior, while others prefer the newer `weight` system. This change makes both workflows available in the same codebase without forcing a full rewrite of inventory-related logic.

---

### Implementation Details

Introduced `Config.InventoryMode` in `es_extended` config, with support for `"weight"` and `"limit"` modes.

Updated inventory loading and item refresh logic to include item `limit` data from the `items` table and propagate it into player inventory objects.

Adjusted `canCarryItem`, `canSwapItem`, and `addInventoryItem` so carrying checks follow the selected inventory mode:
- In `weight` mode, behavior remains based on total carried weight.
- In `limit` mode, behavior checks the target item's configured max count.

Extended the SQL schema for `items` with a `limit` column using `-1` as the default unlimited value, preserving backward-compatible behavior for items without a cap.

Updated inventory type annotations to reflect the added `limit` field.

---

### Usage Example

```lua
Config.InventoryMode = "weight" -- or "limit"
```

---

### PR Checklist


valentinyt20422 (commented):
XD

PR #1780, opened 2026-05-07
fix: mysql.ready called before functions.lua is loaded

FabianTerhorst:
### Description

The code inside `common.lua` is awaiting mysql to be ready. When its ready it is executing RefreshJobs. This function is loaded from inside functions.lua which is below the common.lua inside the manifest. This change fixes the load order and makes sure RefreshJobs is always loaded before common.lua is executed.

### PR Checklist

### DIFF
--- a/[core]/es_extended/fxmanifest.lua
+++ b/[core]/es_extended/fxmanifest.lua
@@ -21,12 +21,13 @@ server_scripts {
 	'@oxmysql/lib/MySQL.lua',
     'shared/config/logs.lua',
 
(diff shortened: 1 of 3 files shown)


Zykem (approved):
Thanks @FabianTerhorst, merging.

Note: this isn't fixing a current bug. It's prep for an upcoming FiveM scheduler change where CreateThreadNow runs on the current stack instead of deferring. Under that, MySQL.ready could fire before functions.lua loads and ESX.RefreshJobs would be nil.

PR #1787, opened 2026-06-11
fix: refactor robustness fixes across core modules

ASTROWwwW:
## What

A batch of **security and robustness fixes** found during an audit of `esx_core` 1.13.5 (still present on `main`). Each change carries an inline `-- Fix:` / `// Fix:` comment. All changes are intentionally **minimal and aim to be non-breaking**; debatable design changes were deliberately left out (see *Not changed* below).

This is a broad PR touching several modules — happy to **split it into per-module PRs** if you prefer to cherry-pick.

## Security

- **`es_extended` — `esx:getOtherPlayerData` (IDOR)**: the callback returns `identifier`, `accounts`, `money`, full `inventory`, live `coords` and `metadata` for a **client-controlled `target`** with no check, so any client could loop IDs `1..N` and exfiltrate every online player's PII + live position. Added a **server-side proximity guard** (source/target peds must be valid and within ~10.0 units) — the least-breaking mitigation that kills remote mass-exfiltration. *You may want to go further and also restrict the returned fields; flagging for discussion.*
- **`esx_multicharacter` — `CharacterChosen` (IDOR / `disablechar` bypass)**: the chosen `charid` came from the client and was loaded without checking it belongs to the player's license or that the character isn't `disabled`. Added server-side validation (slot bound + ownership + `disabled` check). *Adds one `MySQL.single.await` on character selection — worth a test.*
- **`es_extended` — `esx:useItem` / `esx:giveInventoryItem`**: nil-deref on client-controlled item/account names crashed the handler. Added nil/existence guards.
- **`esx_textui`** — message rendered via `innerHTML`; added HTML escaping before color spans are applied (color codes still render).
- **`esx_identity`** — type-guard the registration callback payload, normalize `sex` case.
- **`esx_skin`** — clamp the client-supplied `bags_1` max-weight modifier to config.
- **`es_extended`** — `esx:updateWeaponAmmo` clamps client ammo; `esx:executeCommand` only runs for a non-empty string; `VehicleProperties` statebag is type-guarded.

## Correctness / robustness

- **`removeAccountMoney`** — the underflow guard `balance - money > balance` is always false for `money > 0`, so accounts could go negative. Fixed to `if money > balance` (matches the existing "Tried To Underflow" intent; also affects `removeMoney`).
- **`points` / `point` import** — `SizeOf(points) + 1` reused handles after a removal and silently overwrote an existing point; switched to a monotonic counter. Also fixed a loop-stop using `#` on a sparse table.
- **`cron`** — each job now runs under `pcall`, so one failing job no longer aborts the remaining jobs for that tick.
- **`RefreshJobs`** — guarantee an `unemployed` fallback (an empty jobs table broke player loading).
- **`onesync` `getNearbyPlayers`** — validate `source` before dereferencing; single `GetEntityCoords`.
- **`vehicle:delete`** — stricter entity-validity guard.
- **`SavePlayers`** — duration log used the wrong unit/resolution (always ~0.00).

## Performance

- **`paycheck`** — hoist per-tick constants/translations out of the per-player loop.
- **`interactions`** — removed an unused table that was written every interaction (small leak).
- **`events`** — pickup loop only does proximity work when a pickup is actually active.
- **`actions`** — `Init` made idempotent (was spawning a duplicate tracking thread).
- **`GetRandomString` / `Math.Random`** — seed once at load instead of reseeding with `GetGameTimer()` on every call.

## Not changed (flagged on purpose)

- **`esx:spawnVehicle`** has no permission gate, but many resources call it legitimately — only added defensive arg validation, no permission change.
- **`esx_notify`** uses `RegisterNetEvent` (vs a secure variant) and **`esx_menu_list`** uses a `{{{ }}}` triple-stache that renders generated `<button>` HTML — changing either risks breaking legitimate behavior, so left as-is.

## Testing

Verified the edited regions parse (stock `luac` rejects the pre-existing CfxLua extensions `?.`, `+=`, backtick hashes — unrelated to these edits). Runtime testing recommended, especially `esx_multicharacter` (new DB query) and `paycheck`.

### DIFF
--- a/[core]/cron/server/main.lua
+++ b/[core]/cron/server/main.lua
@@ -38,7 +38,11 @@ function OnTime(timestamp)
 
         if timestamp >= scheduledTimestamp and (not lastTimestamp or lastTimestamp < scheduledTimestamp) then
             local d = os.date('*t', scheduledTimestamp).wday
-            cronJobs[i].cb(d, cronJobs[i].h, cronJobs[i].m)
+            -- Fix: wrap the callback in pcall so one failing job does not interrupt the remaining ones.
+            local ok, err = pcall(cronJobs[i].cb, d, cronJobs[i].h, cronJobs[i].m)
+            if not ok then
+                print(("[^1ERROR^7] cron job (%02d:%02d) raised an error: %s"):format(cronJobs[i].h, cronJobs[i].m, tostring(err)))
+            end
         end
     end
 end
--- a/[core]/es_extended/client/modules/actions.lua
+++ b/[core]/es_extended/client/modules/actions.lua
@@ -226,6 +226,13 @@ function Actions:PedLoop()
 end
 
 function Actions:Init()
+    -- Fix: guard against the double Init() (module load + esx:playerLoaded) that spawned
+    -- a redundant TrackPedCoordsOnce thread and duplicate loops. Init is now idempotent.
+    if self.initialized then
+        return
+    end
+    self.initialized = true
+
     self:SlowLoop()
     self:PedLoop()
     self:TrackPedCoordsOnce()
--- a/[core]/es_extended/client/modules/events.lua
+++ b/[core]/es_extended/client/modules/events.lua
@@ -130,6 +130,11 @@ AddStateBagChangeHandler("VehicleProperties", nil, function(bagName, _, value)
         return
     end
 
+    -- Fix: validate the state bag payload type before use to avoid acting on malformed data
+    if type(value) ~= "table" then
+        return
+    end
+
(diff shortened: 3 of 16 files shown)


Arctos2win:
Hi @ASTROWwwW, nice work. 

I have send you a friend request on discord. 
There is something i wanna ask you.

ASTROWwwW:
> Hi @ASTROWwwW, nice work.
> 
> I have send you a friend request on discord. There is something i wanna ask you.

Hi, thank you.
I didn't receive it. My username is astrowwar

PR #1790, opened 2026-06-17
fix(es_extended/server): harden player save against invalid ped state

ASTROWwwW:
## Summary

Harden server-side player persistence so coordinates and health/armor survive
every save context (disconnect, periodic autosave, resource restart, `/save`,
shutdown) instead of being corrupted to `0,0,0` / `0` when no valid ped is
available at save time (e.g. a player crashing out).

## Changes (server-side only)

- **Coordinates** — `getCoords` (server/classes/player.lua) keeps a canonical
  cached position refreshed on every live read, and returns the last known
  position instead of `0,0,0` when a fresh ped read isn't available.
- **Health / armor** — `updateHealthAndArmorInMetadata` (server/functions.lua)
  reads from a valid ped only, otherwise preserves the last persisted values
  instead of overwriting them with `0`.
- **Save flow** — `Core.SavePlayer` no longer early-returns on `not spawned`,
  so a not-yet-spawned player is saved against cached/loaded values.

## Scope

This PR persists the **position** of a dead player (death location is no longer
lost to `0,0,0`). It does **not** persist or restore the death **state** itself —
es_extended has no server-side notion of death (that lives in the death resource),
so staying dead across reconnect is out of scope here.

## Testing (local ESX server)

- Disconnect → reconnect, alive: respawns at the last persisted position with
  the last persisted HP/armor. Values are the last persisted snapshot, not the
  exact instant of disconnect.
- Disconnect → reconnect, dead: respawns at the last persisted position (death
  location preserved). Death-state restoration is out of scope and unchanged.

## Checklist

- [x] Tested locally
- [x] Conventional commits
- [x] No breaking changes (nominal save path unchanged)

### DIFF
--- a/[core]/es_extended/server/functions.lua
+++ b/[core]/es_extended/server/functions.lua
@@ -190,8 +190,11 @@ end
 
 local function updateHealthAndArmorInMetadata(xPlayer)
     local ped = GetPlayerPed(xPlayer.source)
-    xPlayer.setMeta("health", GetEntityHealth(ped))
-    xPlayer.setMeta("armor", GetPedArmour(ped))
+    -- Fix: only read health/armor from a valid ped; otherwise keep the last saved metadata.
+    if ped ~= 0 and DoesEntityExist(ped) then
+        xPlayer.setMeta("health", GetEntityHealth(ped))
+        xPlayer.setMeta("armor", GetPedArmour(ped))
+    end
     xPlayer.setMeta("lastPlaytime", xPlayer.getPlayTime())
 end
 
--- a/[core]/es_extended/server/functions.lua
+++ b/[core]/es_extended/server/functions.lua
@@ -190,8 +190,8 @@ end
 
 local function updateHealthAndArmorInMetadata(xPlayer)
     local ped = GetPlayerPed(xPlayer.source)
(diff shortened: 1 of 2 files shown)


[core]/es_extended/server/functions.lua
```diff
@@ -190,19 +190,18 @@ end
...
 local function updateHealthAndArmorInMetadata(xPlayer)
     local ped = GetPlayerPed(xPlayer.source)
-    xPlayer.setMeta("health", GetEntityHealth(ped))
-    xPlayer.setMeta("armor", GetPedArmour(ped))
+    -- Fix: only read health/armor from a valid ped, and skip while the player is dead so a persisted health of 0 isn't overwritten with the live ped's HP.
+    if ped ~= 0 and DoesEntityExist(ped) and not Player(xPlayer.source).state.isDead then
+        xPlayer.setMeta("health", GetEntityHealth(ped))
+        xPlayer.setMeta("armor", GetPedArmour(ped))
+    end
     xPlayer.setMeta("lastPlaytime", xPlayer.getPlayTime())
```
Comment (Arctos2win) [inline [core]/es_extended/server/functions.lua:198]: xPlayer.getPlayTime() uses GetPlayerTimeOnline native. This might throw an error if that native returns nil and the method tries to add nil to a number.

PR #1792, opened 2026-06-22
feat(es_extended): grant StartingInventoryItems under ox_inventory

ASTROWwwW:
The ox bridge only seeded StartingAccountMoney, so StartingInventoryItems was ignored under ox. Now granted via ox_inventory:AddItem on new characters, for parity with the default inventory.

Guarded against the false default (pairs(false) would crash).

Not tested at runtime — validation = create a character with StartingInventoryItems set.


Arctos2win (commented):
Starting inventory items are added here for both default inventory and ox inventory:
https://github.com/esx-framework/esx_core/blob/f9a93070c1ed1c207d4992c4e46bff708baf8271/%5Bcore%5D/es_extended/server/main.lua#L39

PR #1796, opened 2026-06-24
feat(esx_lib): add a pub/sub multicast module

ASTROWwwW:
Adds `xLib.pubsub`, a server-driven topic multicast: a producer subscribes
players to a topic server-side and publishes data pushed only to them. Clients
only listen. For things like a shop pushing live stock updates to the players
who have it open, without broadcasting to everyone.

**Server:** `subscribe`, `unsubscribe`, `publish`, `subscribers`, `isSubscribed`, `clear`
**Client:** `on`, `off`

```lua
-- server
xLib.pubsub.subscribe(source, 'myshop:247:stock')
xLib.pubsub.publish('myshop:247:stock', stock)
-- client
xLib.pubsub.on('myshop:247:stock', function(stock) updateUI(stock) end)
```

`subscribe`/`publish` are server-only (never a net event); subscriptions are
cleaned up on `playerDropped` and on resource stop.

### DIFF
--- a/[core]/esx_lib/resource/pubsub/server.lua
+++ b/[core]/esx_lib/resource/pubsub/server.lua
@@ -1,0 +1,207 @@
+-- xLib pub/sub: a producer subscribes players to a topic server-side and
+-- publishes data pushed only to them. Clients only listen, never subscribe.
+-- The registry is shared here (esx_lib server VM). Namespace topics by resource
+-- (e.g. "esx_shops:247:stock") so two resources cannot clash on a bare name.
+
+---@diagnostic disable: duplicate-set-field
+
+local PUBSUB_EVENT <const> = '__xLib_pubsub' -- must match imports/pubsub/client.lua
+local TOPIC_PATTERN <const> = '^[%w_:%.%-]+$'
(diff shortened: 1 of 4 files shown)


[core]/esx_lib/resource/pubsub/server.lua
```diff
@@ -0,0 +1,214 @@
...
+    -- no isValidTopic() here on purpose: an invalid topic never created a set,
+    -- so the lookup below already returns 0 for it without a regex on the hot path
+    local set = subs[topic]
+    if not set then
+        return 0
+    end
+
+    -- no yield in this loop, so the set cannot mutate mid-iteration
+    local count = 0
+    for src in pairs(set) do
```
Comment (Arctos2win) [inline [core]/esx_lib/resource/pubsub/server.lua:108]: Check: 
https://github.com/esx-framework/esx_core/blob/f9a93070c1ed1c207d4992c4e46bff708baf8271/%5Bcore%5D/es_extended/server/functions.lua#L13

( We should add a method in the lib so that its imported rather than sending it to es_extended )

Arctos2win (changes requested):

N0tNvll (approved):

PR #1799, opened 2026-07-05
perf(es_extended): drop redundant player loops in OneSync.SpawnVehicle

ASTROWwwW:
`OneSync.SpawnVehicle` looped over all players twice. Now grabs one via `next(ESX.Players)` for the vehicle-type lookup, and replaces the `GetClosestPlayer` net-owner wait with a plain `DoesEntityExist` check. The server owns the entity right after creation regardless of players in scope, so it still spawns with or without a player nearby.

Tested in-game with `/car`, same behavior.


Zykem (approved):
Tested, LGTM

PR #1809, opened 2026-07-22
fix(esx_lib): load imports from the correct resource name

ASTROWwwW:
### Description

`xLib.name` was `"xLib"`, which is not a resource, so the import metatable fed it to `LoadResourceFile` and resolved nothing: silent on Legacy, a hard error on Enhanced. Points it at the real resource name. Boots clean on both.

### PR Checklist

-   [x] My commit messages and PR title follow the Conventional Commits standard.
-   [x] I have provided a clear explanation of what my PR does.


Zykem (approved):
LGTM

PR #1810, opened 2026-07-22
feat(es_extended): auto-detect Legacy vs Enhanced edition

ASTROWwwW:
### Description

Adds `ESX.IsEnhanced` so one codebase runs on both Legacy and GTA V Enhanced (server: `gamename` convar, client: `IsGameEnhancedVersion`, mirrored to a replicated `esx:enhanced` convar). OneSync Infinity is enforced on Legacy only.

### PR Checklist

### DIFF
--- a/[core]/es_extended/fxmanifest.lua
+++ b/[core]/es_extended/fxmanifest.lua
@@ -14,6 +14,7 @@ shared_scripts {
     'shared/config/adjustments.lua',
 
     'shared/main.lua',
+    'shared/enhanced.lua',
     'shared/functions.lua',
     'shared/modules/*.lua',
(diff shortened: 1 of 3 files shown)


Zykem (changes requested):
this PR is more of an esx_lib thing tbh. the whole point of creating the lib was to keep the core clean and not filled with modules that don't belong there. same logic applies to this variable, it has nothing to do with core functionality, so it should live in the lib instead.

ASTROWwwW:
Agreed, this belongs in esx_lib rather than the core. Moving the detection into the lib and opening a single PR against dev. Thanks for the review.

PR #1813, opened 2026-07-27
fix(es_extended/server/functions): persist items added via AddItems

### LINKED ISSUE #1768: [Bug] - es_extended - Issue ESX.AddItems function
**Describe the bug**
I have encountered a bug with the `ESX.AddItems` function. While the items are correctly added in game, they are not being created in the database (SQL). 

### Steps to Reproduce
1. Use `ESX.AddItems` to create a new item
2. Check the database: the item is missing.
3. Check the in-game inventory: the item is present.

**Debug Info:**
 - OS: Windows
 -  FiveM Artifact windows:master:17000
 - ESX Version: 1.13.4
 - OXMysql : 2.12.0

#### OLD : 

Image

#### NEW : 

Image

seltonmt012:
ESX.AddItems adds the items to ESX.Items but they never end up in the database.

MySQL.prepare batch mode wants an array of positional parameter arrays, but toInsert is a list of keyed tables, so the ? placeholders get nothing and the INSERT silently does nothing. In-game it looks fine because ESX.Items is filled from the same keyed tables, but after a restart the items are gone since they were never saved.

Fixed by building a positional parameters array before the prepared insert.

Tested locally on artifact 25770 with MariaDB: added an item through AddItems and confirmed the row is now written to the items table.

Closes #1768

### DIFF
--- a/[core]/es_extended/server/functions.lua
+++ b/[core]/es_extended/server/functions.lua
@@ -799,9 +799,15 @@ if not Config.CustomInventory then
         end
 
         if #toInsert > 0 then
+            local parameters = {}
+            for i = 1, #toInsert do
+                local row = toInsert[i]
+                parameters[i] = { row.name, row.label, row.weight, row.rare, row.canRemove }
+            end
+
             MySQL.prepare.await(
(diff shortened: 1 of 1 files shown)


seltonmt012:

PR #1815, opened 2026-07-27
fix(es_extended/server/main): use the identity sex for the default skin

seltonmt012:
A character that has no saved skin always falls back to the male model, even when the identity was registered as female.

`loadESXPlayer` builds the fallback skin from `userData.sex`:

```lua
userData.skin = (result.skin and result.skin ~= "") and json.decode(result.skin) or { sex = userData.sex == "f" and 1 or 0 }
```

`userData` is created earlier in the function without a `sex` field, and `userData.sex` is only assigned about thirty lines further down while reading the identity data off the result row. At this point it is still nil, so the comparison is always false and the fallback is always `{ sex = 0 }`.

`result.sex` already holds the value here — it is selected alongside the other identity columns and used a few lines later — so reading it directly is enough.

In practice the character shows up correctly during identity registration, then switches to the male model once the skin data arrives from the server.

Tested locally on artifact 25770 with MariaDB, using a character whose `skin` column is empty:

| identity sex | skin.sex before | skin.sex after |
| --- | --- | --- |
| `f` | 0 (male model) | 1 (female model) |
| `m` | 0 | 0 |

### DIFF
--- a/[core]/es_extended/server/main.lua
+++ b/[core]/es_extended/server/main.lua
@@ -312,7 +312,7 @@ function loadESXPlayer(identifier, playerId, isNew)
     userData.coords = json.decode(result.position) or Config.DefaultSpawns[ESX.Math.Random(1,#Config.DefaultSpawns)]
 
     -- Skin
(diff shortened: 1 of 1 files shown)


seltonmt012:

PR #1817, opened 2026-07-27
fix(es_extended/shared/config): reset the female move speed multiplier to normal

seltonmt012:
`Config.PlayerStatsByGender.female.moveSpeed` is `7.00`, while `male.moveSpeed` is `1.0` and the legend in the same block defines the scale as `1.0 = normal, 1.5 = +50%, 0.5 = -50%`.

The value is applied directly as a ped move rate override in a per frame loop (`client/modules/adjustments.lua`, `SetPedMoveRateOverride`), and `PlayerStatsByGender.enabled` and `useCharacterData` are both true out of the box. So on a stock install every character registered as female moves at seven times the normal rate — fast enough to outrun traffic, and enough to trip the speed checks in most anticheats.

Looks like a leftover test value; `1.0` matches the male entry and the documented scale.

Tested locally on artifact 25770 with a female character: unplayably fast before the change, normal after. A male character behaves the same before and after.

### DIFF
--- a/[core]/es_extended/shared/config/adjustments.lua
+++ b/[core]/es_extended/shared/config/adjustments.lua
@@ -71,7 +71,7 @@ Config.PlayerStatsByGender = {
         stamina = 1.15,
         strength = 0.85,
         swimSpeed = 1.25,
-        moveSpeed = 7.00,
+        moveSpeed = 1.0,
     }
 }
 


seltonmt012:
Closing this one for good. Contributions target `v1.14.1` now, and `Config.PlayerStatsByGender` only exists on `dev` - there is nothing to change on the release branch.

PR #1818, opened 2026-07-27
fix(es_extended/modules/callback): compare the promise state against its numeric value

seltonmt012:
The 15 second timeout in `ESX.AwaitServerCallback` and `ESX.AwaitClientCallback` can never fire.

```lua
SetTimeout(15000, function()
    if p.state == "pending" then
        p:reject("Server Callback Timed Out")
    end
end)
```

`promise.state` is a number, not a string. The runtime defines `PENDING = 0` (through `REJECTED = 4`) and `Citizen.Await` tests `state == 0` itself, so the string comparison is always false. If the other side never answers — handler not registered, an error thrown before `cb()`, a resource restarted mid-call — `Citizen.Await` blocks the calling coroutine permanently instead of rejecting after 15 seconds as the comment above it describes.

Measured on artifact 25770 with the real 15 second delay:

| case | guard | state at 15s | rejected | result of the await |
| --- | --- | --- | --- | --- |
| never answers | current | 0 (pending) | no | never returns |
| never answers | fixed | 0 (pending) | yes | returns "Server Callback Timed Out" |
| answers after 1s | fixed | 3 (resolved) | no | returns normally |

The third row is the regression case: an already resolved promise is left alone, so callbacks that answer in time behave exactly as before.

Worth stating explicitly: after this change a callback that never answers surfaces as a rejection after 15 seconds instead of hanging forever, since `Citizen.Await` re-raises it. That is what the existing comment intends, and nothing can meaningfully depend on the hang. Nothing inside `esx_core` calls either function, so no core behaviour changes.

### DIFF
--- a/[core]/es_extended/client/modules/callback.lua
+++ b/[core]/es_extended/client/modules/callback.lua
@@ -102,7 +102,7 @@ function ESX.AwaitServerCallback(eventName, ...)
 
     -- if the server callback takes longer than 15 seconds to respond, reject the promise
     SetTimeout(15000, function()
-        if p.state == "pending" then
+        if p.state == 0 then
             p:reject("Server Callback Timed Out")
         end
     end)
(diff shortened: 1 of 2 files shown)


seltonmt012:

PR #1841, opened 2026-07-27
fix(es_extended/server/functions): reject invalid weapon arguments instead of raising

seltonmt012:
The `weapon` argument type checks a command argument with `ESX.GetWeapon`, but that function asserts on an unknown name and upper-cases its argument before doing anything else. It can never return a falsy value, so the `else` branch that sets `commanderror_invalidweapon` is dead code, and a typo raises a Lua error in the console instead of telling the admin what was wrong.

Commands registered with `validate = false`, such as `/giveammo`, skip the argument count check and reach `string.upper(nil)`, which raises as well.

Calling it through `pcall` makes both cases take the error branch that was already written for them. Valid names, including lower case, behave as before.

Measured on artifact 25770 over RCON:

```
before:  giveweapon 1 WEAPON_PISTOI 50 -> SCRIPT ERROR ... Invalid weapon name!
         giveammo 1                    -> SCRIPT ERROR ... bad argument #1 to 'upper'
after:   neither errors, both rejected with the invalid weapon message
         giveweapon 1 WEAPON_PISTOL 50 works in both
```

Affects `/giveweapon`, `/giveammo` and `/giveweaponcomponent`.

- [x] Conventional Commits.
- [x] Tested locally.
- [x] No breaking changes.
- [x] Clear explanation.

### DIFF
--- a/[core]/es_extended/server/functions.lua
+++ b/[core]/es_extended/server/functions.lua
@@ -122,8 +122,10 @@ function ESX.RegisterCommand(name, group, cb, allowConsole, suggestion)
                                     err = TranslateCap("commanderror_invaliditem")
                                 end
                             elseif v.type == "weapon" then
-                                if pcall(ESX.GetWeapon, args[k]) then
-                                    newArgs[v.name] = string.upper(args[k])
+                                local weapon = ESX.GetWeaponFromHash(args[k])
+
+                                if weapon then
+                                    newArgs[v.name] = weapon.name
                                 else
                                     err = TranslateCap("commanderror_invalidweapon")
                                 end
--- a/[core]/es_extended/server/functions.lua
+++ b/[core]/es_extended/server/functions.lua
@@ -122,7 +122,7 @@ function ESX.RegisterCommand(name, group, cb, allowConsole, suggestion)
                                     err = TranslateCap("commanderror_invaliditem")
                                 end
                             elseif v.type == "weapon" then
-                                if ESX.GetWeapon(args[k]) then
+                                if pcall(ESX.GetWeapon, args[k]) then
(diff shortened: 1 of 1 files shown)


N0tNvll:
Is not solving the problem continue returning true, fix it

seltonmt012:
Fair point, the pcall hides the raise instead of avoiding it. Replaced it with `ESX.GetWeaponFromHash`, which already returns nil for a name it does not know. joaat is case insensitive and a nil key is a legal table read, so a typo and a missing argument both come back as nil and land in the `commanderror_invalidweapon` branch that was already written. The accepted value now comes from the config entry instead of upper casing the raw argument.

Headless over RCON, on a command that takes a single weapon argument:

```
                 before          after
WEAPON_PISTOL    accepted        accepted
weapon_pistol    accepted        accepted
WEAPON_PISTOI    SCRIPT ERROR    Invalid weapon
no argument      SCRIPT ERROR    Invalid weapon
```

A hash looking argument such as `453432689` is rejected as well.

In game with one player connected: `/giveweapon 1 WEAPON_PISTOI 50` and `/giveweaponcomponent 1 WEAPON_PISTOI clip_extended` both answer with the error message, `/giveweapon 1 weapon_pistol 50` gives the weapon, and the `/giveammo 1 WEAPON_PISTOL 30` after it goes through, which only happens when the player really has it. `/giveammo 1` answers with the error instead of the `string.upper(nil)` crash.

I left `ESX.GetWeapon` itself alone. Dropping the assert changes its contract, and the three core callers index the second return value right away, so they would fail one line later with a worse message. Say the word and I'll do that as its own PR.

PR #1843, opened 2026-07-27
fix(esx_lib/streaming): give asset requests a timeout

seltonmt012:
Every request function loops `while not HasXLoaded(...) do Wait(500) end` and never gives up. A misspelled or unavailable anim dict, texture dict or ptfx asset makes that loop run forever, so the calling coroutine never returns. `esx_progressbar` waits on `RequestAnimDict`, so a bad dict there leaves the player stuck. A not yet loaded asset also costs at least 500 ms, since the first check always fails.

`dev` fixed this in 0f5f6a6b with a `waitForLoaded` helper: 5000 ms timeout, 50 ms step, and a shortcut when the asset is already loaded. It was not carried over when the module moved into `esx_lib`, so `v1.14.1` and `main` still run the old code. This ports it and does the same for `requestAudioBank`, which `dev` does not have.

Callers already deal with nil. `requestModel` could always return nil through the `IsModelInCdimage` guard, and `esx_multicharacter` and `spawnVehicle` both check for it.

Measured in game on artifact 25770 with a dict name that does not exist:

```
before:  never returned, still hanging after 8 seconds
after:   returns nil after 5026 ms
```

A valid dict still returns immediately.

- [x] Conventional Commits.
- [x] Tested locally.
- [x] No breaking changes.
- [x] Clear explanation.

### DIFF
--- a/[core]/esx_lib/imports/streaming/client.lua
+++ b/[core]/esx_lib/imports/streaming/client.lua
@@ -34,8 +34,9 @@ end
 
 ---@param modelHash number | string
 ---@param cb? function
+---@param timeout? number milliseconds to wait, defaults to 5000
 ---@return number | nil
-xLib.streaming.requestModel = function(modelHash, cb)
+xLib.streaming.requestModel = function(modelHash, cb, timeout)
     modelHash = type(modelHash) == "number" and modelHash or joaat(modelHash)
 
     if not IsModelInCdimage(modelHash) then return end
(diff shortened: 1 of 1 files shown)


N0tNvll:
I suggest implementing a default streaming timeout and allowing an optional timeout for each streaming function

seltonmt012:
Added. The default stays at 5000 ms and every request function takes an optional timeout after the callback, so existing calls are unchanged.

Measured in game on artifact 25770:

```
bad anim dict, no timeout arg        nil after 5004 ms
bad anim dict, 1000 ms               nil after 1040 ms
bad anim dict, 200 ms                nil after  231 ms
bad texture dict, 800 ms             nil after  830 ms
valid anim dict, no timeout arg      returned after   0 ms
valid anim dict, 1000 ms             returned after  56 ms
valid anim dict, callback, 3000 ms   callback result after 0 ms
valid model, 4000 ms                 returned after 119 ms
```

PR #1854, opened 2026-07-27
fix(es_extended/server/modules/createJob): let CreateJob add grades to an existing job

seltonmt012:
### Description

`ESX.CreateJob` always builds an `INSERT INTO jobs` regardless of whether the job is already there. `jobs` has its primary key on `name`, so as soon as the row exists the transaction fails on a duplicate key and rolls back, and nothing gets written at all.

The guard above it does not stop that. It only returns early when the job **and every** grade you passed already exist, which is exactly the case where you have nothing to add.

---

### Motivation

Adding a grade to an existing job is the obvious use for this function and it cannot work today:

| | before | after |
| --- | --- | --- |
| `police` has grades 0-3, add grade 4 | `Duplicate entry 'police' for key 'PRIMARY'`, nothing written | grades 0-3-4 |
| `jobs` row exists with no grades, add grade 0 | same error, nothing written | grade 0 added |
| brand new job | works | unchanged |
| job and grade both already there | early return with a message | same message, same early return |

The second row is reachable without anyone doing anything unusual. `ESX.RefreshJobs` drops any job with no grades out of `ESX.Jobs` and logs `Ignoring job "x" due to no job grades found`, but the `jobs` row stays in the database. The guard only looks at `ESX.Jobs`, sees nothing, and lets it through into the duplicate insert. That job can then never be repaired through `CreateJob`.

The error message points the wrong way too. `transaction_async` returns `false`, so the caller is told `Failed to insert one or more grades`, when the grades were fine and the `jobs` row was the problem.

---

### Implementation Details

Ask the database what is already there, then only insert what is missing:

```lua
local jobExists = MySQL.scalar.await('SELECT 1 FROM `jobs` WHERE `name` = ?', { name }) ~= nil
```

Checking the database rather than `ESX.Jobs` is what covers the grade-less job case. The old guard is gone because `if not queries[1]` now covers the same situation and gives the same message, and it no longer depends on `ESX.Jobs` being in sync.

`generateNewJobTable` starts from the existing entry when there is one. Without that, adding grade 4 to `police` would leave `ESX.Jobs["police"].grades` holding only grade 4 and drop the other four out of memory until the next restart.

Two notes on the grade side. `job_grades` has its primary key on `id`, not on `job_name, grade`, so a repeated grade does not raise anything, it just writes a second row. `existingGrades` stops that. And the old `doesJobAndGradesExist` used `ipairs` while the insert loop uses `pairs`, so a grades table with string keys was never seen by the guard at all.

Checked against a scratch database built from the schema in `[SQL]/legacy.sql`, running the old and the new statement sequence over four cases: grade added to an existing job, grade added to a job with no grades, a brand new job, and nothing to add. The last two behave the same before and after.

---

### Usage Example

```lua
ESX.CreateJob("police", "LSPD", {
    { grade = 4, name = "boss", label = "Captain", salary = 100 },
}, "leo")

-- before: [ERROR] Failed to insert one or more grades for job: `police`
--         and the grade is not there
-- after:  [SUCCESS] Job created successfully: `police`
```

---

### PR Checklist

### DIFF
--- a/[core]/es_extended/server/modules/createJob.lua
+++ b/[core]/es_extended/server/modules/createJob.lua
@@ -52,14 +52,17 @@ function ESX.CreateJob(name, label, grades, jobType)
         jobType = "civ"
     end
 
-    local jobExists = MySQL.scalar.await('SELECT 1 FROM `jobs` WHERE `name` = ?', { name }) ~= nil
+    local existingJob = MySQL.single.await('SELECT `label`, `type` FROM `jobs` WHERE `name` = ?', { name })
+    local jobExists = existingJob ~= nil
     local existingGrades = {}
 
     if jobExists then
+        label, jobType = existingJob.label, existingJob.type
+
         local rows = MySQL.query.await('SELECT `grade` FROM `job_grades` WHERE `job_name` = ?', { name })
 
         for i = 1, #(rows or {}) do
-            existingGrades[rows[i].grade] = true
+            existingGrades[tostring(rows[i].grade)] = true
         end
     end
 
@@ -72,8 +75,11 @@ function ESX.CreateJob(name, label, grades, jobType)
         }
     end
 
+    local newGrades = {}
+
     for _, grade in pairs(grades) do
-        if not existingGrades[grade.grade] then
+        if not existingGrades[tostring(grade.grade)] then
+            newGrades[#newGrades + 1] = grade
             queries[#queries + 1] = {
                 query = 'INSERT INTO job_grades (job_name, grade, name, label, salary, skin_male, skin_female) VALUES (?, ?, ?, ?, ?, ?, ?)',
                 values = { name, grade.grade, grade.name, grade.label, grade.salary, grade.skin_male and json.encode(grade.skin_male) or '{}', grade.skin_female and json.encode(grade.skin_female) or '{}' }
@@ -93,7 +99,7 @@ function ESX.CreateJob(name, label, grades, jobType)
         return success
     end
 
-    ESX.Jobs[name] = generateNewJobTable(name, label, grades, jobType)
+    ESX.Jobs[name] = generateNewJobTable(name, label, newGrades, jobType)
 
     notify("SUCCESS", currentResourceName, 'Job created successfully: `%s`', name)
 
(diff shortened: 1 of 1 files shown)


N0tNvll:
generateNewJobTable() still receives the full grades table, including grades that already exist in the database fix that cause will overwrite the info

seltonmt012:
That was real, thanks. Reproduced it on a running server: create the job with grades 0 and 1, then call `CreateJob` again with grade 0 carrying different data plus a new grade 2.

```
after: DB  grade 0 = rookie / Rookie / 100    row untouched
after: MEM grade 0 = CHANGED / CHANGED / 999  overwritten
```

The row stayed as it was and `ESX.Jobs` carried the caller's values until the next restart. `salary` is the one that hurts, it reaches the paycheck through `setJob`. `generateNewJobTable` now only receives the grades that were actually inserted.

Two things came out of it while I was in the file:

* **Label and type of an existing job now come from the database.** `RefreshJobs` drops jobs without grades from `ESX.Jobs` (`functions.lua:556`), so a job can sit in the table without being in memory, and `ESX.Jobs[name] or {...}` would have cached the caller's label over the one the row holds. Measured with `DB Label`/`police` in the row against `Passed Label`/`civ` passed in: memory takes the row, and no second `jobs` row is written.
* **The grade lookup compares as a string.** `job_grades` has no unique key on `(job_name, grade)`, so a grade passed as `"1"` would not have matched the number `1` coming back from the query and a duplicate row would have gone in. The old `DoesJobExist` compared with `tostring`, so that tolerance was there before.

Third commit is a separate bug in the same function. `generateNewJobTable` put the caller's skin table into `ESX.Jobs` while the row got the encoded string, so `setJob` ran `json.decode` over a table:

```
before: MEM skin_male type=table  -> json.decode ok=false, bad argument #1 to 'json.decode' (string expected, got table)
after:  MEM skin_male type=string -> json.decode ok=true, identical to the row
```

A restart made the job look healthy again, because the grade then came back from the database as a string. The cached value is now the same one that goes into the row.

One limit worth naming: I drove all of this through a command built into `es_extended` itself, not through a third party resource calling `ESX.CreateJob`.

PR #1875, opened 2026-09-01
Update loadscreen file path in fxmanifest.lua

N0tNvll:
### Description

Redundant code in fxmanifest

---

### Motivation

---

### Implementation Details

---

### Usage Example

```lua
-- Add example usage here
```

---

### PR Checklist

-   [ ] My commit messages and PR title follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard.
-   [ ] My changes have been tested locally and function as expected.
-   [ ] I have provided a clear explanation of what my PR does, including the reasoning behind the changes and any relevant context.

### DIFF
--- a/[core]/esx_loadingscreen/fxmanifest.lua
+++ b/[core]/esx_loadingscreen/fxmanifest.lua
@@ -8,10 +8,6 @@ description 'Allows resources to Run tasks at specific intervals.'
 version '1.15.0'
 lua54 'yes'
 
-loadscreen 'index.html'
-
-shared_script 'config.lua'
-
 loadscreen "web/index.html"
 loadscreen_manual_shutdown "yes"
 


ASTROWwwW (approved):
Fixes the report, the first loadscreen entry pointed to a deleted file. Two things left: the manifest still has the old header block duplicated (two fx_version lines, cron description), and with loadscreen_manual_shutdown the shutdown callback never reaches the resource because GetParentResourceName() inside a loadscreen frame returns the literal frame name 'loadingScreen' (see root.html in the client ui, the frame is created under that name). So the screen will show now but nothing ever calls ShutdownLoadingScreenNui. I have a follow-up ready that hardcodes the callback url and adds a session-start failsafe.

RodericAguilar:
Good

